PR #13, opened 2024-04-03
Add the GNU Radio plug-in into the Lime Suite NG source tree

DominykasPetke:
# Added
- The GNU Radio plug-in in the source tree. 
  - Requires GNU Radio 3.10.4 or newer to build together with everything else.
    - Disabled by default, enable with `-DENABLE_GNURADIO=1` when invoking CMake.
  - Can go to the plug-in's root directory to build it standalone.
  - Supports GNU Radio versions 3.9, 3.10 and current 3.11-git.
- Building the GNU Radio plug-in in GitHub Actions.
- CMake option to not install the kernel driver, only build it.


plugins/gr-limesdr/docs/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
...
+
+########################################################################
+# Setup dependencies
+########################################################################
+find_package(Doxygen)
+
+########################################################################
+# Begin conditional configuration
+########################################################################
+if(ENABLE_DOXYGEN)
```
rjonaitis, line 9: Shouldn't this also check if the Doxygen was actually found?

plugins/gr-limesdr/README.md
```diff
@@ -0,0 +1,37 @@
...
+## Documentation
+
+* [MyriadRF Wiki page](https://wiki.myriadrf.org/Gr-limesdr_Plugin_for_GNURadio)
+
+## Dependencies
+ 
+* GNU Radio (versions 3.9, 3.10 or 3.11)
+* libfmt
+* pybind11
+* LimeSuite
```
rjonaitis: limesuiteng

plugins/gr-limesdr/lib/device_handler.cc
```diff
@@ -0,0 +1,753 @@
...
+        list_read = true;
+    }
+
+    if (serial.empty()) {
+        GR_LOG_INFO(d_logger,
+                    "device_handler::open_device(): no serial number. Using first device "
+                    "in the list.");
+        GR_LOG_INFO(
+            d_logger,
+            "Use \"LimeUtil --find\" in terminal to find preferred device serial.");
```
rjonaitis: LimeUtil is not part of limesuiteng. Now device listing is done using 'limeDevice'

plugins/gr-limesdr/lib/device_handler.cc
```diff
@@ -0,0 +1,753 @@
...
+    antenna_rx[0] = device_handler::getInstance()
+                        .get_device(device_number)
+                        ->GetAntenna(0, lime::TRXDir::Rx, 0);
+
+    if (pAntenna_tx != nullptr) {
+        pAntenna_tx[0] = antenna_tx[0];
+        pAntenna_tx[1] = antenna_tx[1];
+    }
+
+    device_handler::getInstance()
```
rjonaitis: why not just store "device_handler::getInstance()" into local variable once at the start of the function, and then reuse it? Same with `device_handler::getInstance().get_device(device_number)`, it would greatly reduce spam and improve readablility.

plugins/gr-limesdr/lib/device_handler.cc
```diff
@@ -0,0 +1,753 @@
...
+    GR_LOG_DEBUG(d_debug_logger, "device_handler::calibrate(): ");
+
+    double rf_freq = device_handler::getInstance()
+                         .get_device(device_number)
+                         ->GetFrequency(0, direction, channel);
+    if (rf_freq > 31e6) // Normal calibration
+        device_handler::getInstance()
+            .get_device(device_number)
+            ->Calibrate(0, direction, channel, bandwidth);
+    else { // Workaround
```
rjonaitis: technically this is not a valid workaround and should not be performed, as the calibration results at different frequencies are incompatible.

plugins/gr-limesdr/lib/device_handler.cc
```diff
@@ -0,0 +1,753 @@
...
+                               lime::TRXDir direction,
+                               int channel,
+                               double bandwidth)
+{
+    GR_LOG_DEBUG(d_debug_logger, "device_handler::calibrate(): ");
+
+    double rf_freq = device_handler::getInstance()
+                         .get_device(device_number)
+                         ->GetFrequency(0, direction, channel);
+    if (rf_freq > 31e6) // Normal calibration
```
rjonaitis: just execute the calibration without any workarounds, edge cases depending on hardware should be handled inside limesuiteng procedures themselves.

plugins/gr-limesdr/lib/device_handler.cc
```diff
@@ -0,0 +1,753 @@
...
+    if (device_handler::getInstance()
+            .get_device(device_number)
+            ->SetAntenna(0, direction, channel, antenna) != lime::OpStatus::SUCCESS)
+        device_handler::getInstance().error(device_number);
+
+    int antenna_value = device_handler::getInstance()
+                            .get_device(device_number)
+                            ->GetAntenna(0, direction, channel);
+
+    const std::array<std::array<std::string, 4>, 2> s_antenna = {
```
rjonaitis: antenna names can be retrieved from SDRDevice descriptor.

plugins/gr-limesdr/lib/logging.cc
```diff
@@ -0,0 +1,80 @@
...
+    default:
+        break;
+    }
+}
+
+void set_limesuite_logger(void)
+{
+    std::lock_guard<std::mutex> lock(log_handler_mutex);
+    if (!logger)
+        gr::configure_default_loggers(logger, debug_logger, "LimeSuite");
```
rjonaitis: LimeSuiteNG

plugins/gr-limesdr/CMakeLists.txt
```diff
@@ -1 +1,194 @@
...
+   set(CMAKE_BUILD_TYPE "Release")
+   message(STATUS "Build type not specified: defaulting to release.")
+endif(NOT CMAKE_BUILD_TYPE)
+set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "")
+
+# Make sure our local CMake Modules path comes first
+list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_SOURCE_DIR}/cmake/Modules)
+
+# Set the version information here
+set(VERSION_MAJOR 1)
```
rjonaitis, line 33: Is this version used anywhere? Does it have to match the change log `gr-limesdr (4.0.0-0) unstable; urgency=low`

rjonaitis (changes requested):

rjonaitis:
Have you actually tested this module with the included gnu radio companion examples?
Attempting to use the examples, some of the block parameters are invalid, and running the graph shows other problems like:
```
LimeSuite :error: Rx NCO255 index invalid
LimeSuite :warning: Coefficient 18349483025486831236718453623221193645217020343443292089619258547898772706562168057367043541141359895248896,000000 is outside of range [-1:1], incorrect value will be written.
```

DominykasPetke:
> Have you actually tested this module with the included gnu radio companion examples?

Yes, I have and I am able to listen to radio with the FM receiver example.

As for the logs, I am looking into the issues.

plugins/gr-limesdr/docs/CMakeLists.txt
DominykasPetke: It does so in the main file (`plugins/gr-limesdr/CMakeLists.txt`:174), so removed the extraneous `find_package`

plugins/gr-limesdr/CMakeLists.txt
DominykasPetke: This eventually ends up as the `.so` version of the library files.

DominykasPetke: And no, it does not need to match

PR #14, opened 2024-04-05
Add HDSDR plug-in to source tree

DominykasPetke:
# Added
- HDSDR plug-in to the source tree
  - GitHub Action to build it
  
 Based from https://github.com/myriadrf/ExtIO_LimeSDR


CMakeLists.txt
```diff
@@ -137,6 +137,11 @@ if(CMAKE_COMPILER_IS_GNUCXX)
 endif()
 
 if (MSVC)
+    # Check if using correct Visual Studio version
+    if(MSVC_TOOLSET_VERSION LESS 141)
+	    message(FATAL_ERROR "MSVC version should be atleast v141 (Visual Studio 2017)")
```
rjonaitis, line 143: What causes this requirement?
Add a comment why lesser versions can't be used.

rjonaitis: If that is requirement of the HDSDR, than it should only be imposed for HDSDR CMake, and the plugin compilation should be made optional.

plugins/HDSDR/src/ExtIO_LimeSDR.h
```diff
@@ -0,0 +1,69 @@
...
+#define EXTIO_API __declspec(dllexport) __stdcall
+#else
+#define EXTIO_API __declspec(dllimport)
+#endif
+
+#include "LC_ExtIO_Types.h"
+
+#define ExtIO_LimeSDR_VERSION @ExtIO_LimeSDR_VERSION @
+
+extern "C" bool EXTIO_API InitHW(char* name, char* model, int& type);
```
rjonaitis, line 34: extern "C" can be a scope, no need to add it to each individual function declaration.

plugins/CMakeLists.txt
```diff
@@ -26,4 +26,9 @@ option(ENABLE_GNURADIO "Enable build library" OFF)
...
 if (ENABLE_GNURADIO)
     add_subdirectory(gr-limesdr)
-endif()
\ No newline at end of file
+endif()
+
+########################################################################
+## HDSDR plugin
+########################################################################
+add_subdirectory(HDSDR)
```
rjonaitis: make it optional.

.github/workflows/cmake.yml
```diff
@@ -41,7 +45,7 @@ jobs:
         run: brew install libusb wxmac
 
       - name: Download windows dependencies
-        if: matrix.os == 'windows-latest'
+        if: matrix.os == 'windows-latest' && matrix.arch == 'x64'
```
rjonaitis: This is wrong, all dependencies are available for both x64 and x86.
Also this would result in the Win32 build (not installing FX3) and the library would end up without USB support.
Because HDSDR application is Windows x86 only, then just create a separate gitaction build script for it with all it's limitations.

plugins/HDSDR/CMakeLists.txt
```diff
@@ -0,0 +1,117 @@
...
+cmake_minimum_required (VERSION 3.12)
+
+# Only compile on Windows
+if(NOT MSVC)
+	return()
+endif()
+
+# Only compile if 32-bit
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+	return()
```
rjonaitis: Add a message that HDSDR application is available only in 32bits only, so the plugin must be compiled for 32bit architecture as well.

plugins/HDSDR/src/ExtIO_LimeSDR.cpp
```diff
@@ -0,0 +1,1265 @@
...
+    numberOfChannels = descriptor.rfSOC[0].channelCount;
+    if (numberOfChannels <= 0) {
+        return false;
+    }
+
+    if (device->Init() != lime::OpStatus::SUCCESS) {
+        return false;
+    }
+
+    if (device->SetSampleRate(0, lime::TRXDir::Rx, 0, sampleRates.at(sampleRateIndex), oversample) != lime::OpStatus::SUCCESS) {
```
rjonaitis: It seems InitializeLMS() does all parameters setup, so it would be preferrable to use SDRDevice::Configure() function instead of the individual ones, as it guarantees that parameters are set in a specific order and validates if all parameters are valid.

rjonaitis (changes requested):

PR #15, opened 2024-04-05
Improvements to tests

DominykasPetke:
# Changed
- Device tests to actually connect to the device and receive test pattern data
- Updated GTest

# Fixed
- Clang code coverage generation
- A few bugs in USB communications code


src/tests/boards/LimeSDR/LimeSDRTest.cpp
```diff
@@ -1,107 +1,213 @@
...
-using ::testing::AllOf;
-using ::testing::AnyNumber;
-using ::testing::DoAll;
-using ::testing::ReturnArg;
-using ::testing::SetArrayArgument;
+using namespace std::literals::string_literals;
+
+MATCHER_P2(AreSamplesCorrect, samples, maxSampleCount, "Checks if the test pattern gave the correct samples"s)
+{
+    // Samples can begin at any point in the sequence, so we're finding a common start point first.
```
rjonaitis: This strategy skips data verification before the "common" point, so it will pass even if the starting data would be invalid.
The correct strategy to test the "test signal", would be to take the first received complex sample, and generate the expected sequence from that point.
The test sequence is sinusoidal wave sampled at configured time intervals:
So if the TestSignal::Divide is Div4, the sampling phase difference will be (2pi / 4). That makes the next sample in the sequence to be the previous complex(i,q) rotated by (2pi/4).

rjonaitis (changes requested):
All these tests are identical, so it shouldn't be device specific. This is more like a runtime tests of the API.

Device specific testing, that tests actual hardware if it functions properly will not be done using GTest, it will be a standalone application.

DominykasPetke:
I would prefer to still have these tests exist here for the time being (until the mentioned standalone testing application is working) just to have some assurance that devices are actually being connected to and function properly in a consistent, non-GUI reliant manner.

src/tests/boards/LimeSDR/LimeSDRTest.cpp
DominykasPetke: Got it to work, but it seems like it needs a 1-2 bit (in 12-bit space, so 16 or 32 in 16-bit space) delta allowance for it to account for floating point inaccuracies properly.

PR #21, opened 2024-04-18
Clean up string usage

DominykasPetke:
# Added
- Missing `std::string` versions to `Logger.h`

# Changed
- All logging callbacks to use `std::string`
- Switched `std::string` to `std::string view` in select places
- Using `std::string` and `std::string_view` literals where logical
- Some `char*` to `std::string_view` and/or `std::filesystem::path`
- Made sure all links use HTTPS wherever possible

# Removed
- Unnecessary logging functions in LMS7002M


src/boards/LimeSDR_XTRX/LimeSDR_XTRX.cpp
```diff
@@ -133,12 +135,14 @@ LimeSDR_XTRX::LimeSDR_XTRX(std::shared_ptr<IComms> spiRFsoc,
...
-    desc.spiSlaveIds = { { "LMS7002M", SPI_LMS7002M }, { "FPGA", SPI_FPGA } };
+    desc.spiSlaveIds = { { "LMS7002M"s, SPI_LMS7002M }, { "FPGA"s, SPI_FPGA } };
+
+    const std::unordered_map<eMemoryRegion, Region> eepromMap = { { eMemoryRegion::VCTCXO_DAC, { 16, 2 } } };
 
     // desc.memoryDevices[MEMORY_DEVICES_TEXT.at(eMemoryDevice::FPGA_RAM)] =
     //     std::make_shared<DataStorage>(this, eMemoryDevice::FPGA_RAM);
-    const std::unordered_map<eMemoryRegion, Region> flashMap = { { eMemoryRegion::VCTCXO_DAC, { 16, 2 } } };
     desc.memoryDevices[ToString(eMemoryDevice::FPGA_FLASH)] = std::make_shared<DataStorage>(this, eMemoryDevice::FPGA_FLASH);
+    desc.memoryDevices[ToString(eMemoryDevice::EEPROM)] = std::make_shared<DataStorage>(this, eMemoryDevice::EEPROM);
```
rjonaitis: XTRX board does not have EEPROM memory. Values that would be written to non volatile memory, are being stored on the last page of fpga's FLASH memory. The 'flashMap' is the description of byte meanings on that last page.

src/logger/LoggerCString.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+void log(const LogLevel level, const char* format, ...)
+{
+    va_list args;
+    va_start(args, format);
+    log(level, format, args);
+    va_end(args);
+}
+
+constexpr std::size_t MAX_MSG_LEN = 1024;
```
rjonaitis: should be static.

src/logger/Logger.cpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    log(LogLevel::Info, text);
+}
+
+void debug(const std::string& text)
+{
+    log(LogLevel::Debug, text);
+}
+
+//! Log a message with formatting and specified logging level
```
rjonaitis: this just takes a string, not a formatting string with variale arguments.

rjonaitis (changes requested):

PR #23, opened 2024-04-22
Clean up macros

DominykasPetke:
# Removed
- `LMS7param` macro
- Some other smaller miscellaneous macros

# Changed
- Some macros to `constexpr` constants


rjonaitis (changes requested):
Looks good, but the LMS7param macro removal should be reverted from this PR.
There is already work in progress, of this macro is being removed and the LMS7Parameter is being moved into namespace, the changes will land in couple of days. This removal would cause conflicts everywhere where parameters are used.

PR #27, opened 2024-04-29
Fix GNU Radio in-tree building

DominykasPetke:
Running `find_package(Gnuradio)` outside of the `gr-limesdr` project scope permanently sets a few values, which, later on, when trying to actually build the package, looks for files in the wrong place and the compilation fails.

### DIFF
--- a/plugins/gr-limesdr/CMakeLists.txt
+++ b/plugins/gr-limesdr/CMakeLists.txt
@@ -15,8 +15,8 @@ enable_testing()
 
 find_package(Gnuradio QUIET) # precheck if any gnuradio package is available
 include(CMakeDependentOption)
-cmake_dependent_option(
-    ENABLE_GNURADIO "Build LimeSuiteNG plugin for GNU Radio" ON "ENABLE_LIBRARY;Gnuradio_FOUND;BUILD_SHARED_LIBS" OFF)
+include(FeatureSummary)
+cmake_dependent_option(ENABLE_GNURADIO "Build LimeSuiteNG plugin for GNU Radio" ON "ENABLE_GNURADIO_PREREQ;Gnuradio_FOUND" OFF)
 add_feature_info(GNURADIO ENABLE_GNURADIO "GNU Radio 3.9+ Plug-in")
 if(NOT ENABLE_GNURADIO)
     return()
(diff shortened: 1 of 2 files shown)


plugins/gr-limesdr/CMakeLists.txt
```diff
@@ -13,6 +13,15 @@ cmake_minimum_required(VERSION 3.12)
 project(gr-limesdr CXX C)
 enable_testing()
 
+find_package(Gnuradio QUIET) # precheck if any gnuradio package is available
+include(CMakeDependentOption)
+cmake_dependent_option(
+    ENABLE_GNURADIO "Build LimeSuiteNG plugin for GNU Radio" ON "ENABLE_LIBRARY;Gnuradio_FOUND;BUILD_SHARED_LIBS" OFF)
```
rjonaitis: ENABLE_LIBRARY variable would not be available in the standalone build of the plugin, so the ENABLE_GNURADIO would always result in false and would build nothing.

rjonaitis:
What values are being set wrong by the 'find_packge(GnuRadio)'? it should not matter where it was called from, the returned paths are absolute, not relative.

DominykasPetke:
During the `find_package` command a bunch of GNU Radio's CMake files are ran, and at the bottom of `gnuradio/GrPython.cmake` a file is written with `${PROJECT_BINARY_DIR}` as the base path, then, when in the `gr-limesdr` project, when trying to execute the file, `PROJECT_BINARY_DIR` is different, thus the file is unable to be found. The file does not get written again when trying to find the package again because at the top of the `gnuradio/GrPython.cmake` file there is, essentially, an include guard, which, when trying to add the file twice, simply turns the whole operation into a no-op.

rjonaitis:
Does the problem require specific steps to reproduce?
I've had no issues with building the plugin as part of the whole project, and as standalone as well.
I don't see any references of `python_compile_helper` usage, where would it fail?. Is it a remnant of compatibility with old gnu radio versions?
Using current gnuradio's 'gr_modtool' to generate a fresh plugin module, it does not contain 'GrPython.cmake'

DominykasPetke:
Just running a clean CMake/Make run. Not sure from exactly which GNU Radio version it starts happening, but tested with GNU Radio versions 3.10.5.1 and 3.10.9.2, both have this exact behaviour.

And the problem is not in the module itself, but in GNU Radio's CMake files that get executed when running `find_package(Gnuradio)`

PR #31, opened 2024-05-08
Rewrite and improve CLI argument parsing

CoffeeLexer:
- Replaced C UNIX/POSIX getopt with C++ library [`args`](https://github.com/Taywee/args)

### DIFF
--- a/src/cli/limeSPI.cpp
+++ b/src/cli/limeSPI.cpp
@@ -127,7 +127,7 @@ static std::string ReadFile(const std::string& fileName)
 int main(int argc, char** argv)
 {
     // clang-format off
-    args::ArgumentParser                    parser("limeSPI - CLI Memory device flashing", "");
+    args::ArgumentParser                    parser("limeSPI - Control status registers I/O", "");
     args::HelpFlag                          help(parser, "help", "This help", {'h', "help"});
 
     args::Group                             commands(parser, "commands");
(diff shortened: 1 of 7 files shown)


src/cli/limeSPI.cpp
```diff
@@ -103,75 +104,70 @@ static int parseReadInput(std::string_view hexstr, std::vector<uint32_t>& mosi)
...
+    if (!inputFile.is_open())
+    {
+        cerr << "Failed to open file: "sv << fileName << endl;
+        exit(EXIT_FAILURE);
+    }
+    inputFile.seekg(0, std::ios::end);
+    long fileSize = inputFile.tellg();
+    inputFile.seekg(0, std::ios::beg);
+
+    buffer.resize(fileSize + 1); // +1 to add null termination
```
rjonaitis: null termination is not required, as it's not a C-string

src/cli/limeTRX.cpp
```diff
@@ -276,186 +229,95 @@ class ConstellationPlotter
...
-    std::filesystem::path txFilename{ ""sv };
-    bool rx = true;
-    bool tx = false;
-    bool showFFT = false;
+    // clang-format off
+    args::ArgumentParser                parser("limeTRX - Realtime streaming of RF samples", "");
+    args::HelpFlag                      helpFlag(parser, "help", "This help", {'h', "help"});
+
+    args::ValueFlag<std::string>        deviceFlag(parser, "name", "Specifies which device to use", {'d', "device"});
+    args::NargsValueFlag<int>           chipFlag(parser, "index", "Specify chip index, or index list for aggregation [0,1...]", {'c', "chip"}, {1, static_cast<size_t>(-1)});
```
rjonaitis: It's unclear what `{1, static_cast<size_t>(-1)}` is, is it default value, or accaptable range?

rjonaitis (changes requested):
limeTRX has regressed and is not working.

PR #32, opened 2024-05-10
Add a DMA USB wrapper

DominykasPetke:
# Added
- USB DMA wrapper
- A DMA interface class

# Changed
- Merged PCIe TRX Looper code with the main looper code and made it based on the DMA interface
- USB devices use the singular remaining TRX Looper
- Made `LitePCIe` inherit the DMA interface
- Split interfaces in IComms into their own separate files
- Simplified a few pieces of code for ease of reading

# Fixed
- USB enumeration not filtering devices properly on Windows
- `basicTX` example using the wrong device

# Removed
- USB-specific TRX Looper


DominykasPetke:
Putting this in as a draft for now as I feel like a fresh pair of eyes would be quite beneficial at this point, as this does have a workaround that should not be in the final code but have no place to put it in a clean way, as well as probably a few misunderstandings on how DMA actually works, and a few issues I do not have a clear idea on how to fix, and probably some issues I don't even know about

src/tests/protocols/BufferInterleavingTest.cpp
```diff
@@ -6,11 +6,16 @@
 
 using namespace lime;
 
+template<typename... Ts> constexpr std::array<std::byte, sizeof...(Ts)> make_bytes(const Ts&&... args) noexcept
```
rjonaitis: PR is suppose to be focused about USB streaming wrapped with DMA interface. Changes like this are unnecessary. Also adding std::byte, it does not add any benefits in this situation and just overcomplicates things, now this needs variadic templates just to initialize the arrays.

DominykasPetke:
@rjonaitis Not sure if you've seen this

src/protocols/TRXLooper.cpp
```diff
@@ -254,6 +334,287 @@ void TRXLooper::Stop()
...
+    const int requestSamplesInPkt = 4080 / sampleSize / chCount;
+
+    const int payloadSize = requestSamplesInPkt * sampleSize * chCount;
+    const int samplesInPkt = payloadSize / (sampleSize * chCount);
+
+    uint32_t packetSize = payloadSize + headerSize;
+
+#if defined(__unix__) && !defined(__APPLE__)
+    // TODO: fix workaround
+    // only if PCIe device
```
rjonaitis: Variable Rx packet size feature is determined by the device gateware, so this should be made as a capability flag or function in the FPGA class, which could be enabled by the device when creating the fpga object.

src/comms/USB/USBDMA.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    if (samplesDirection == TRXDir::Rx && dataDirection == DataTransferDirection::DeviceToHost)
+    {
+        return true;
+    }
+
+    return false;
+}
+
+void USBDMA::CacheFlush(TRXDir samplesDirection, DataTransferDirection dataDirection, uint16_t index)
```
rjonaitis: USB does not have DMA capability, so it does not need to instruct CPU caches flushing, because this is just only immitating DMA interface, flushing should do nothing.
Flushing is only required when PCIe DMA is used, as the CPU and device is sharing RAM access, the CPU may have cached some of the RAM memory values and is not aware that peripheral device may have modified it at some point.

src/comms/USB/USBDMA.cpp
```diff
@@ -0,0 +1,323 @@
...
+                GetEndpointAddress(direction)));
+        rx.state.hardwareIndex++; // TODO: deal with potential overflow?
+        lck.unlock();
+        std::this_thread::yield();
+    }
+
+    port->AbortEndpointXfers(GetEndpointAddress(direction));
+}
+
+void USBDMA::TxStartTransferThread()
```
rjonaitis: Dedicated transmitting thread is not required.
You can initiate Tx BeginDataXfer when buffer values have been filled and then DMA state is being updated with counter increment.

src/comms/USB/USBDMA.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    const auto received{ port->FinishDataXfer(reinterpret_cast<uint8_t*>(address), GetBufferSize(), contextHandle) };
+
+    if (received != GetBufferSize())
+    {
+        throw std::runtime_error("Did not transfer all bytes"s);
+    }
+}
+
+void USBDMA::RxStartTransferThread()
```
rjonaitis: Dedicated Rx thread is also not required.
When RxDMA is enabled, request the initial batch of transfers. Then request next transfer, when Rx updates the DMA state with incremented software counter.

rjonaitis (changes requested):

PR #33, opened 2024-05-15
Rewrite driver device path management

CoffeeLexer:
- Fixed udev-rules being not applied
- Renamed litepcie to limelitepcie
- Changed all devices to use generic names
- Grouped device specific communication files into directories
- Added version data to Linux PCIe Device driver

Device Hierarchy after changes:
- Old
  /dev/limex30_control
  /dev/limex30_trx0
- New
  /dev/limelitepcie0/control0
  /dev/limelitepcie0/trx0
  /dev/limelitepcie0/trx1


9600:
Why "limelitepcie? Have we forked and modified LitePCIe? How do we benefit even from an SDR Linux device entry having "litepcie" in the naming? It feels needlessly long, but I'm probably missing something.

rjonaitis:
the name "limelitepcie" can be changed to anything, it just needs to be unique, suggestions are welcome. This choice was just a straigthforward "lime" prefix addition to the original "litepcie" name, as it is a modified version of it.
I'm not sure about the hardware side, but I think it is using the original litepcie as it is. On the software side, the driver code has been modified to our needs and forked from the original, to provide functionality specific to our devices that the original 'litepcie' does not have (like: device specific DMA channels count, and buffer sizes, other device VIDs/PIDs, specific device naming).

There are couple of reasons for the rename:
1. The kernel module name itself. The name has to be unique, to not clash with the original "litepcie" driver if it were already installed in the system, and both could be present at the same time.
2. The system device names in the `/dev` directory. Human readable and obvious names like "LimeSDR****" are convenient, but unwieldy. As that would require driver code update (just adding additional name string) for each new device released, even though the functionality of the driver remains the same. Also in case old drivers would detect newer devices by VID/PID but not have a human readable name assigned in that version, that would still require to have some kind of fallback name for such cases, or for the driver to ignore that device, and bring confusion why it is working.

The "/dev/limelitepcie*" nameing was chosen to be the same as the module name for consistency sake. It can also be changed to anything we want, it doesn't have to match the module name, suggestions are welcome. The requirement is that it needs to be unique from other system device names. As this generic device name is intended for any of our possible PCI devices, it wouldn't be technically correct to include SDR in the name, as it's functionality could be anything else. Just to clarify these names are for the system usage to allow it to work (like: sda, sdb, tty0, tty1, i2c-0, i2c-1.... they just indicate type of device and index ), human readable symlink names can be provided additionaly by using UDEV rules, or list the detailed info about devices using our CLI tools `limeDevice`

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1226,53 +1208,78 @@ static pci_ers_result_t litepcie_error_detected(struct pci_dev *dev, pci_channel
...
 };
 
-static int litepcie_alloc_chdev(struct litepcie_device *s)
+static void device_create_release(struct device *dev)
+{
+    pr_debug("device: '%s': %s\n", dev_name(dev), __func__);
+    kfree(dev);
+}
+
+struct device *lime_device_create(const struct class *class, struct device *parent, dev_t devt, void *drvdata, const char *s)
```
rjonaitis: unused function?

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1226,53 +1208,78 @@ static pci_ers_result_t litepcie_error_detected(struct pci_dev *dev, pci_channel
...
-    .write = litepcie_ctrl_write,
+    .unlocked_ioctl = limelitepcie_ioctl_control,
+    .open = limelitepcie_ctrl_open,
+    .release = limelitepcie_ctrl_release,
+    .read = limelitepcie_ctrl_read,
+    .write = limelitepcie_ctrl_write,
 };
 
-static int litepcie_alloc_chdev(struct litepcie_device *s)
+static void device_create_release(struct device *dev)
```
rjonaitis: unused?

src/comms/PCIe/linux-kernel-module/boards.h
```diff
@@ -38,7 +38,7 @@ enum eLMS_DEV {
 };
 
 const char LMS_DEV_NAMES[][80] = {
-    "UNKNOWN",
+    "LimeSDR", // Unknown
```
rjonaitis: Signify that it is recognized as LimeSDR, but unknown "LimeSDR-Unknown"

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1475,7 +1483,11 @@ static int ReadInfo(struct litepcie_device *s)
     }
     s->info.serialNumber = serialNumber;
 
-    snprintf(s->info.devName, sizeof(s->info.devName), "%s", GetDeviceName(s->info.boardId));
+    if (0 < s->info.boardId && s->info.boardId < LMS_DEV_COUNT)
```
rjonaitis: GetDeviceName already handles invalid values, returns unknown device name.

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1487,26 +1499,54 @@ static int ReadInfo(struct litepcie_device *s)
...
+        return true;
+    return false;
+}
+
+static bool ValidChar(char c)
+{
+    return isalnum(c) || c == '-' || c == '_' || c == ' ';
+}
+
+static void ParseIdentifier(char *identifier, char *destination)
```
rjonaitis: identifier should be const.

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1487,26 +1499,54 @@ static int ReadInfo(struct litepcie_device *s)
...
+{
+    uint8_t size = 0;
+    while (size < 256 && ValidChar(identifier[size]))
```
rjonaitis: 256 should be replaced with identifier length.

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1487,26 +1499,54 @@ static int ReadInfo(struct litepcie_device *s)
...
+    {
+        if (identifier[size] == ' ')
+            identifier[size] = '-';
+        size++;
+    }
+
+    while (size > 0 && identifier[size - 1] == '-')
+        size--;
+
+    strncpy(destination, identifier, size);
```
rjonaitis: destination max length should be checked to not overflow the buffer.

src/comms/PCIe/linux-kernel-module/limelitepcie.c
```diff
@@ -1565,16 +1607,18 @@ static int litepcie_pci_probe(struct pci_dev *dev, const struct pci_device_id *i
...
 
-    ret = AllocateIRQs(litepcie_dev);
+    ret = AllocateIRQs(limelitepcie_dev);
     if (ret < 0)
         goto fail2;
 
-    if (ReadInfo(litepcie_dev) != 0)
+    if (ValidIdentifier(fpga_identifier))
+        ParseIdentifier(fpga_identifier, limelitepcie_dev->info.devName);
+    else if (ReadInfo(limelitepcie_dev) != 0)
```
rjonaitis: log an error message if the ReadInfo fails.

rjonaitis (changes requested):

src/comms/PCIe/LimeLitePCIe.cpp
```diff
@@ -35,47 +36,66 @@ std::vector<std::string> LitePCIe::GetDevicesWithPattern(const std::string& rege
     return devices;
 }
 
-std::vector<std::string> LitePCIe::GetPCIeDeviceList()
+std::vector<std::string> LimeLitePCIe::GetPCIeDeviceList()
 {
     std::vector<std::string> devices;
     FILE* lsPipe;
-    lsPipe = popen("ls -1 -- /sys/class/litepcie", "r");
+    lsPipe = popen("ls -1 -- /sys/class/limelitepcie", "r");
```
rjonaitis: if there is no pcie devices found by the driver, the /sys/class/limelitepcie directory will not exist and the "ls" command will print a error "ls: cannot access '/sys/class/limelitepcie': No such file or directory", the err stream should be discarded.

rjonaitis (approved):

PR #35, opened 2024-05-23
Port LMS7002M code to C

DominykasPetke:
# Changed
- Ported some LMS7002M code to C
  Things currently not ported:
  - Everything that uses files
  - Functions that heavily use `LMS7002M_RegistersMap`
  - Functions that use the MCU
  
# Removed
- Calibration code that used the MCU for calibration
- Other unused bits of code in LMS7002M class


src/chips/LMS7002M/calibrations.cpp
```diff
@@ -1,1539 +0,0 @@
...
-            lime::error(messageFormat, signalLevel_dBFS, expectedSignalLevel_dBFS);
-            return MCU_LOOPBACK_SIGNAL_WEAK;
-        }
-        else
-            lime::warning(messageFormat, signalLevel_dBFS, expectedSignalLevel_dBFS);
-    }
-    return MCU_NO_ERROR;
-}
-
-uint8_t CalibrateRx(bool extLoopback, bool dcOnly)
```
rjonaitis, line 1423: This file contains the actual calibration code, which should be executed. It shoud not be removed.

rjonaitis (changes requested):
`MCU_BD` port is not needed, and it is not required for calibrations. All it does, is send parameters to the MCU inside LMS7002M and delegate the calibration procedure to it.
The MCU should not be used for calibration anymore. The `src/chips/LMS7002M/calibrations.cpp` should be ported to C as well, and the calibration should be performed by the host system itself.

rjonaitis:
[Convert floats to doubles](https://github.com/myriadrf/LimeSuiteNG/pull/35/commits/58c4f4e9b3189a9cf084cfb868d624a3eb19e183)
The embedded directory code is intended to be used by micro controllers, they don't always have a native double precision, resulting in silent downgrade to single precision, or generating slow emulated implementation of double. Use only `float` precision in this code.

embedded/lms7002m/lms7002m.c
```diff
@@ -0,0 +1,1886 @@
...
+            txLpfHighRange[0]);
+        powerDowns = 0x07;
+    }
+    else if (rfBandwidth_Hz <= txLpfHighRange[1]) // <160MHz
+    {
+        const float Hlog = 20 * log10(rfbandwidth_MHz / (28 * 2));
+        float Hterm1;
+        {
+            float t1 = 5.66735e-16;
+            float t2 = sqrt(1.21443429517649e103 * pow(Hlog, 2) - 2.85279160551735e104 * Hlog + 1.72772373636442e105);
```
rjonaitis: these have to be double precision, otherwise the operations end up producing 'NaN' or 'Inf'

embedded/lms7002m/lms7002m.c
```diff
@@ -0,0 +1,1886 @@
...
+        lms7002m_log(self, lime_LogLevel_Warning, "TxLPF(%g) setting bandwidth to %g.", rfBandwidth_Hz, txLpfLowRange[0]);
+        powerDowns = 0x11;
+    }
+    else if (rfBandwidth_Hz <= txLpfLowRange[1]) // 5.3-33 MHz
+    {
+        const float LADlog = 20.0 * log10(rfbandwidth_MHz / (2.6 * 2));
+        float LADterm1 = 0.0;
+        {
+            float t1 = 1.92163e-15;
+            float t2 = sqrt(5.9304678933309e99 * pow(LADlog, 2) - 1.64373265875744e101 * LADlog + 1.17784161390406e102);
```
rjonaitis: must be doubles

embedded/lms7002m/save_state.c
```diff
@@ -0,0 +1,80 @@
...
+uint16_t chipStateData[359]; // UPDATE IF SOMETHING CHANGES
+
+void lms7002m_save_chip_state(lms7002m_context* self, bool wr)
+{
+    uint16_t dest = 0;
+    const uint16_t ch = lms7002m_spi_read(self, 0x0020);
+    if (!wr)
+        x0020state = ch;
+
+    for (uint8_t i = 0; i < sizeof(chipStateAddr) / sizeof(uint16_t*); ++i)
```
rjonaitis: This is wrong, the loop value is incorrect, as the multidimensional array is not consisting of pointers. It fails to process the entire state.

rjonaitis (changes requested):

rjonaitis:
Rx channel B calibration is failing, investigating why.

PR #38, opened 2024-05-31
Simplify some functions

DominykasPetke:
# Changed
- Simplified `ILMS7002MTab` device parameter read/write functions
- Simplified `Si5351C` code massively
- Replaced iterator magic in `MCU_File.cpp` with functions from the `std::` library
- Simplified multiple functions in CDCM6208 code
- Simplify a loop in MCU_BD panel code
- Deduplicated `GenerateWindowCoefficients` function


src/chips/LMS7002M/MCU_File.cpp
```diff
@@ -425,32 +426,25 @@ void MCU_File::ReadHex(unsigned long limit)
     {
         throw std::runtime_error("No data in file!"s);
     }
-    vector<MemBlock>::iterator vi;
-    m_top = 0;
-    for (vi = m_chunks.begin(); vi < m_chunks.end(); vi++)
-    {
-        m_top = std::max(m_top, (vi->m_startAddress + vi->m_bytes.size() - 1));
```
rjonaitis, line 432: This is max value search, not accumulation.

src/chips/Si5351C/Si5351C.cpp
```diff
@@ -44,483 +44,58 @@ void realToFrac(const float real, int& A, int& B, int& C)
...
+    { 91, 0x1E },
+    { 181, 0x30 },
+    { 183, 0xD2 },
+    { 184, 0x60 },
+    { 185, 0x60 },
+    { 187, 0xC0 },
+    { 221, 0x0D },
+} };
+
+static_assert(defaultConfiguration.back() != Si5351C_Memory{ 0, 0 },
```
rjonaitis: How is this assertion related to the amount of elements in the array?

src/comms/IComms.h
```diff
@@ -80,45 +80,45 @@ class IComms : public ISPI
       @param bufLength The length of @p buffer.
       @return The operation success state.
      */
-    virtual OpStatus GPIOWrite(const uint8_t* buffer, const size_t bufLength) { return OpStatus::NotImplemented; };
+    virtual OpStatus GPIOWrite(const uint8_t* buffer, const size_t bufLength) = 0;
```
rjonaitis: Why? what's the benefit of this?
IComms is abstract class of what operations could be performed, it's not a requirement that all methods need to have an implementation. So by default all of them are reporting that they are not implemented.

The side effect of this change is that now every class that provides only a part of this interface, has to explicitly add overrides for all other methods just to indicate that, resulting in a lot of useless code spam.

rjonaitis (changes requested):

src/chips/Si5351C/Si5351C.cpp
DominykasPetke: When there are less elements specified in the array than the set size of the array, the remaining values are default constructed (in this case to `{0, 0}`), so the assertion fails (assuming the last intended element is not meant to be `{0, 0}`). If there are more elements to be constructed in the array than the array can hold, then there is just a normal compilation error about there being not enough space for all the elements.

src/chips/LMS7002M/MCU_File.cpp
DominykasPetke: It is a max value search, but since it is not a straight forward maximum element search (which could easily be done with `std::max_element`), using this, I feel like, is the cleanest solution.

Code comparison between the original, current and a third, potential idea:
https://godbolt.org/z/8sr8qr3MP

src/comms/IComms.h
DominykasPetke: Original idea was to just a nice clean interface for [C++ Core Guidelines point C.121](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c121-if-a-base-class-is-used-as-an-interface-make-it-a-pure-abstract-class), however, reverting this to not have much spamminess.

src/chips/LMS7002M/MCU_File.cpp
rjonaitis: There is no need to overcomplicate things and use every available feature of the language.
A simple loop is far more comprehensible and readable than the lambda monstrosities.
```
struct MemBlock {
    size_t m_startAddress;
    std::vector<unsigned char> m_bytes;
    size_t GetEndAddress() const { return m_startAddress + m_bytes.size() - 1; }
};

m_highestAddress = 0;
for (const auto& chunk: m_chunks)
    m_highestAddress = std::max(m_highestAddress, chunk.GetEndAddress());
```

src/chips/Si5351C/Si5351C.cpp
rjonaitis: just use const vector, it won't require this management overhead. This is not a performance critical code, so no need to micro optimize it with fixed sized array.

PR #39, opened 2024-05-31
Add `clang-tidy` with a base of checks under the `cppcoreguidelines` name scheme

DominykasPetke:
# Added
- `clang-tidy` configuration files
- GitHub Action to run `clang-tidy` and error out on anything that should be corrected

# Fixed
- Issues raised by `clang-tidy`

# Removed
- Some unused class members and functions


plugins/SoapyLMS7/Settings.cpp
```diff
@@ -28,14 +28,14 @@ using namespace lime;
  ******************************************************************/
 SoapyLMS7::SoapyLMS7(const DeviceHandle& handle, const SoapySDR::Kwargs& args)
     : _moduleName(handle.media)
+    , sdrDevice([&] {
```
rjonaitis: why use lambda? this just makes it unreadable mess and no real benefit in terms of efficiency.

plugins/gr-limesdr/lib/device_handler.cc
```diff
@@ -606,13 +607,16 @@ void device_handler::set_nco(int device_number,
 
     const std::array<std::string, 2> s_cmix = { "UPCONVERT"s, "DOWNCONVERT"s };
 
-    std::array<double, 16> freq_value_out;
-    double pho_value_out;
+    double pho_value_out{ NAN };
+    const auto freq_value_out{ [&] {
```
rjonaitis: stop with the lambdas, it serves no purpose here, just makes everything less readable.

src/protocols/LMS64CProtocol.cpp
```diff
@@ -33,9 +33,12 @@ using namespace lime::LMS64CProtocol;
 namespace lime {
 
 LMS64CPacket::LMS64CPacket()
+    : cmd(Command::GET_INFO)
+    , status(CommandStatus::Undefined)
 {
     static_assert(sizeof(LMS64CPacket) == 64);
-    std::memset(this, 0, sizeof(LMS64CPacket));
```
rjonaitis, line 38: This was pretty clear and simple.
Now the structure data initialization is split among three places: header, constructor initializer list, and constructor implementation. Technically valid, but more confusing what is initialized where.

rjonaitis (changes requested):

plugins/gr-limesdr/lib/device_handler.cc
DominykasPetke: [C++ Core Guidelines point ES.28](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es28-use-lambdas-for-complex-initialization-especially-of-const-variables), and as this is essentially a `const` variable, it's fitting.

However, cleaning this up as only the 0th index value was ever used in the log afterwards anyway.

plugins/SoapyLMS7/Settings.cpp
DominykasPetke: `clang-tidy` wants the initialization there, so only real options are:
* removing the log
* log gets written post-connection
* adding a `NOLINT` comment so `clang-tidy` is quiet about the initialization not being in the constructor initializer list
* moving the log together with the initialization into the lambda (chosen option)

rjonaitis: There is another option, disable the rule that enforces this.
They suppose to be guidlines, not actual rules. Plus the C++ core guidelines are targeted for C++17,20 and up, we're not targeting that. Even though the project standard is set to C++17, it's to allow some features, not force every feature usage.
In general, blindly enforcing rules such as this is counter productive when they are not necessary. Is it more efficient, technically correct: yes, does it make it easier to read/debug and follow instructions: no. Prefer efficiency in performance critical code, but leave better readability in places like this where the code is executed once.

DominykasPetke: The rule itself that does this ([`cppcoreguidelines-prefer-member-initializer`](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/prefer-member-initializer.html)) is readability *and* performance focused, as just scrolling through this PR alone it shows its improvements to readability, and, arguably, this is the only place where it does indeed negatively affect readability or does behaviour changing changes. Thus I'll just leave a `NOLINT` comment here and hopefully that is satisfactory enough.

PR #54, opened 2024-06-17
Set up DKMS for the Limepcie kernel module

DominykasPetke:
# Added 
- DKMS support for the kernel module


rjonaitis (changes requested):
This assumes that DKMS is always available on the system. It fails if dkms is not present, and there is no other way left to build the module. 
Subsequent "make install" fails to update the kernel module, as dkms install fails because module with the same version already is available.

The build system should provide option whether to use DKMS, or simply compile and load the module as before.

PR #55, opened 2024-06-17
GNU Octave plug-in upgrade

DominykasPetke:
# Added
- Re-enabled GNU Octave plug-in compilation
- Added `liboctave` dependency to Linux GitHub Action build
- `EnableTxWaveform` to `SDRDevice`

# Changed
- Ported the GNU Octave plug-in to the new API
- GNU Octave API changes:
  - `LimeReceiveSamples` does not take a channel parameter anymore, instead return all the samples at once (in a form of a `(channels, samples)` shaped matrix)
  - `LimeTransmitSamples` does not take a channel parameter anymore, instead takes all the samples at once (in a form of a `(channels, samples)` shaped matrix)
  - `LimeTransceiveSamples` - same deal as the above functions


eabase:
This look pretty useful! Any reason why it hasn't been merged?

PR #62, opened 2024-07-02
Update external dependencies

DominykasPetke:
# Changed
- Now downloading dependencies on CMake set-up instead of having them inside the repository
- Upgraded internal `pll_sweep` to new API

# Updated
- `cpp-feather-ini-parser` [from `4299cbc6` to `c04a5e50`](https://github.com/c-ridgway/cpp-feather-ini-parser/compare/4299cbc683ba0a9ca38904dc60d77b379e801230...c04a5e504f7b954548fbbd763ac0d4843d4d7c93)
- `kissFFT` [from `a3a09b39` to `f5f2a3b2`](https://github.com/mborgerding/kissfft/compare/a3a09b3946dde21a31d6f3965b63e57c37f7e5a0...f5f2a3b2f2cd02bf80639adb12cbeed125bdf420)
- GLEW from 1.13.0 to 2.2.0
- GTest to the latest main

# Added
- File denoting which version of FTD3XXLibrary is currently in-use

# Removed
- `getopt.h`, as we have moved on to use `args`


rjonaitis (changes requested):
-- Install configuration: "Release"
-- Up-to-date: /usr/local/lib/libkissfft-float.so.131.1.0
-- Up-to-date: /usr/local/lib/libkissfft-float.so.131
-- Up-to-date: /usr/local/lib/libkissfft-float.so
-- Up-to-date: /usr/local/include/kissfft/kiss_fft.h
-- Up-to-date: /usr/local/include/kissfft/kissfft.hh
-- Up-to-date: /usr/local/include/kissfft/kiss_fftnd.h
-- Up-to-date: /usr/local/include/kissfft/kiss_fftndr.h
-- Up-to-date: /usr/local/include/kissfft/kiss_fftr.h
-- Up-to-date: /usr/local/lib/cmake/kissfft/kissfft-float-shared-targets.cmake
-- Up-to-date: /usr/local/lib/cmake/kissfft/kissfft-float-shared-targets-release.cmake
-- Up-to-date: /usr/local/lib/cmake/kissfft/kissfft-config.cmake
-- Up-to-date: /usr/local/lib/cmake/kissfft/kissfft-config-version.cmake
-- Up-to-date: /usr/local/lib/pkgconfig/args.pc

using `fetchcontent_makeavailable` makes external packages as standalone projects and adds them to the install/uninstall procedure, which is not good. It means that any random changeset that is specified to be fetched for the external repo will be installed on the host system, potentially overriding already existing files and changing their versions. Same for uninstall, it will delete those files even if they were originally installed prior to this project. This kind of side effect should be avoided.

9600:
Does running cmake trigger dependency source download? If so, I suspect package builders such as Ubuntu Launchpad/PPA won't be happy with this. All dependencies should be resolved by the OS and where dependencies cannot be met by the distro and e.g. adding other PPAs / repos, as a last resort we will have to package the dependency. 

Our build shouldn't trigger downloading additional sources. 

Or maybe I've misunderstood.

rjonaitis:
compiling/build can download external dependencies, but they all should be linked statically.

9600:
Package builder policy aside - maybe it's not an issue with Launchpad, but can't help but wonder more generally - it intuitively doesn't feel great and like it potentially makes build more fragile.

PR #65, opened 2024-07-02
TRXLooper: reduce spinlocking

DominykasPetke:
# Changed
- Removed spinlocks in TRXLooper and replaced them with condition variables


rjonaitis:
Did you test?
Attempting to stop Tx stream freezes application.

src/protocols/TRXLooper.cpp
```diff
@@ -339,8 +342,9 @@ OpStatus TRXLooper::RxSetup()
 #endif
 
     // wait for Rx thread to be ready
+    std::unique_lock lck{ mTx.mutex };
     while (mRx.stage.load(std::memory_order_relaxed) < Stream::ReadyStage::WorkerReady)
-        ;
+        mTx.cv.wait(lck);
```
rjonaitis, line 347: Again, poorly tested. Wrong conditional variable, causes lock up. Fixed in 'develop'

PR #66, opened 2024-07-02
Copy wxWidgets DLLs to the binary on build

DominykasPetke:
# Added
- Automatically copy over wxWidgets DLLs on Windows to the binary directory to remove the need to manually copy them.

# Changed
- Updated wxWidgets used on Windows to 3.2.5

### DIFF
--- a/GUI/CMakeLists.txt
+++ b/GUI/CMakeLists.txt
@@ -11,6 +11,13 @@ if(NOT wxWidgets_FOUND AND WIN32)
     cmake_policy(SET CMP0135 NEW)
     message(STATUS "Downloading wxWidgets headers.")
     set(wxRoot ${PROJECT_BINARY_DIR}/_deps/wxWidgets)
+
+    # If 2017 or 2019
+    if(MSVC_TOOLSET_VERSION LESS 143)
+        set(additional_folder "/${CMAKE_BUILD_TYPE_INIT}")
+        set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE_INIT})
+    endif()
+
     if(CMAKE_SIZEOF_VOID_P EQUAL 8)
         fetchcontent_declare(
             wxWidgets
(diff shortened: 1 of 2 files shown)


GUI/CMakeLists.txt
```diff
@@ -11,27 +11,52 @@ if(NOT wxWidgets_FOUND AND WIN32)
     cmake_policy(SET CMP0135 NEW)
     message(STATUS "Downloading wxWidgets headers.")
     set(wxRoot ${PROJECT_BINARY_DIR}/_deps/wxWidgets)
-    if(CMAKE_CL_64)
+
+    # If 2017 or 2019
```
rjonaitis: Comment why this is needed.

rjonaitis (changes requested):
Doesn't work. Visual Studio 2022, dlls get copied into /bin directory. While the actual generated binaries are in /bin/Debug(or Release).

DominykasPetke:
I am unable to reproduce VS2022 creating the additional /Debug (or /Release) directory across two different VMs.

Yes, that directory gets created when using 2017 or 2019 build tools, but not with the 2022 ones.

Versions being used:
VS Build Tools 2017: 15.9.63
VS Build Tools 2019: 16.11.37
VS Community 2022: 17.10.3

rjonaitis:
with VS Community 2022: 17.10.3 I'm still getting Debug/Release directories.

By design MSVC is multi configuration, so it would always create Debug/Release subdirectories
[RUNTIME_OUTPUT_DIRECTORY_CONFIG](https://cmake.org/cmake/help/latest/prop_tgt/RUNTIME_OUTPUT_DIRECTORY_CONFIG.html#prop_tgt:RUNTIME_OUTPUT_DIRECTORY_%3CCONFIG%3E)
Multi-configuration generators ([Visual Studio](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#visual-studio-generators), [Xcode](https://cmake.org/cmake/help/latest/generator/Xcode.html#generator:Xcode), [Ninja Multi-Config](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html#generator:Ninja%20Multi-Config)) append a per-configuration subdirectory to the specified directory unless a generator expression is used.

Also CMAKE_BUILD_TYPE_INIT is not really documented and is internal CMAKE variable, I think it could be replaced with using generator expressions.

PR #69, opened 2024-07-02
Enable GTest on shared library builds

DominykasPetke:
# Fixed
- Enabled GTest on shared library builds
- Cleaned up a few tests


src/protocols/BufferInterleaving.h
```diff
@@ -11,8 +12,8 @@ struct DataConversion {
     uint8_t channelCount;
 };
 
-int Deinterleave(void* const* dest, const uint8_t* buffer, uint32_t length, const DataConversion& fmt);
-int Interleave(uint8_t* dest, const void* const* src, uint32_t count, const DataConversion& fmt);
+int LIME_API Deinterleave(void* const* dest, const uint8_t* buffer, uint32_t length, const DataConversion& fmt);
```
rjonaitis: Requiring to export all symbols just for the sake of testing them is not good, as they will be left exported even without testing enabled.
If testing is enabled, just conditionally compile in all the tests to the library itself, that way tests can access all the internal functions. And the tests can be stripped out when releasing.

rjonaitis (changes requested):

PR #70, opened 2024-07-02
Additional Doxygen documentation

DominykasPetke:
# Added
- More Doxygen documentation

# Changed
- Renamed `USBGeneric` to `UnixUsb` to better reflect what the class actually is
- Some class members' accessibility levels.

# Fixed
- Some misdocumented parts.
- OEMTesting crashes.


src/DSP/FFT/FFT.cpp
```diff
@@ -71,99 +37,6 @@ void FFT::ConvertToDBFS(std::vector<float>& bins)
...
-    kiss_fft(m_fftCalcPlan, m_fftCalcIn.data(), m_fftCalcOut.data());
-    outputBins.resize(count);
-    for (uint32_t i = 0; i < count; ++i)
-    {
-        float amplitude = (m_fftCalcOut[i].r * m_fftCalcOut[i].r + m_fftCalcOut[i].i * m_fftCalcOut[i].i) / (count * count);
-        outputBins[i] = amplitude > 0 ? 10 * log10(amplitude) : -150;
-    }
-}
-
-void FFT::ProcessLoop()
```
rjonaitis: This functionality should not be removed, it is intended to be reusable replacement for logic hardcoded into GUI
https://github.com/Matom-ai/LimeSuiteNG/blob/9e86b01433e66c04c2697158f2fac7c36c588476/GUI/fftviewer_wxgui/fftviewer_frFFTviewer.cpp#L529-L604

rjonaitis (changes requested):

PR #80, opened 2024-07-05
Use the FFT class in FFTViewer

DominykasPetke:
# Changed
- Now using FFT class in FFTViewer, thus moving some of the logic away from the GUI.


GUI/fftviewer_wxgui/fftviewer_frFFTviewer.cpp
```diff
@@ -491,169 +468,126 @@ void fftviewer_frFFTviewer::StreamingLoop(
...
-                    if (samplesToCopy <= 0)
-                        break;
-                    for (uint32_t i = 0; i < samplesToCopy; ++i)
-                    {
-                        captureBuffer[ch][samplesCaptured + i].real(buffers[ch][i].real() * 32767);
-                        captureBuffer[ch][samplesCaptured + i].imag(buffers[ch][i].imag() * 32767);
-                    }
-                    samplesToCapture -= samplesToCopy;
-                    samplesCaptured += samplesToCopy;
+                    pthis->fftBins[channel][i] = bins[i];
```
rjonaitis: can just memcpy.

rjonaitis (changes requested):
Data channels are intented to be synced, they are received and displayed together, so it makes more sense to add multi channel version of the FFT, than to try to split the received data into two independent FFT threads and then sync them back together for display.

src/DSP/FFT/RingBuffer.h
```diff
@@ -52,34 +68,61 @@ template<class T> class RingBuffer
...
+        }
         tailIndex += count;
         size += count;
         produced += count;
         return produced;
     }
 
+    /// @brief Gets the amount of items in the buffer.
+    /// @return The amount of items currently in the buffer.
+    std::size_t Size() const { return size; }
```
rjonaitis: size, is neither atomic, nor guarded by lock, creates possibility for race condition.

src/DSP/FFT/RingBuffer.h
```diff
@@ -52,34 +68,61 @@ template<class T> class RingBuffer
...
     {
         std::lock_guard<std::mutex> lck(mMutex);
         count = std::min(count, capacity - size);
-        int produced = 0;
+        std::size_t produced = 0;
 
         if (tailIndex + count >= capacity)
         {
-            int toCopy = capacity - tailIndex;
-            memcpy(&buffer[tailIndex], src, toCopy * sizeof(T));
```
rjonaitis, line 64: if data type T is_trivially_copyable, use memcpy.

rjonaitis (changes requested):

src/DSP/FFT/FFT.cpp
```diff
@@ -1,17 +1,22 @@
...
 #include <chrono>
 
 namespace lime {
 
-FFT::FFT(uint32_t size)
-    : samplesFIFO(size * 128)
+FFT::FFT(uint8_t channelCount, uint32_t size, WindowFunctionType windowType)
+    : samplesFIFO(channelCount)
+    , channelCount(channelCount)
+    , currentWindowType(windowType)
```
rjonaitis, line 12: only sets the window type, but does not generate coefficients for it, and default to GenerateWindowCoefficients(WindowFunctionType::BLACKMAN_HARRIS), until next type change, resulting in GUI selection and actual value mismatch.

PR #81, opened 2024-07-05
Allow specifying antenna for basicRX

sergeyvfx:
This change adds rxpath command line argument which can be used to configure the antenna used for reception. Additionally, the change makes it so configured antenna is printed.

### DIFF
--- a/src/examples/basicRX.cpp
+++ b/src/examples/basicRX.cpp
@@ -136,7 +136,7 @@ int main(int argc, char** argv)
     config.channel[0].tx.enabled = false;
     config.channel[0].tx.sampleRate = sampleRate;
     config.channel[0].tx.oversample = 2;
-    config.channel[0].tx.path = rxPath;
+    config.channel[0].tx.path = 0;
     config.channel[0].tx.centerFrequency = frequencyLO - 1e6;
     config.channel[0].tx.testSignal.enabled = false;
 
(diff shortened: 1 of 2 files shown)


sergeyvfx:
Such functionality was very useful for some investigation on the LimeSDR XTRX and Mini I am doing, and I saw there is a TODO in the code about it. Wasn't sure that was the intended way to solve it, but figured since I have the code for it already, why not to provide it.
Let me know if that is desired direction, which will also be good to have for the `basicTX.cpp`.
If you had something else in mind, then I don't mind either adjusting the code, or if you'd like to tackle it on your end and close this PR.

src/examples/basicRX.cpp
```diff
@@ -63,21 +100,43 @@ int main(int argc, char** argv)
...
-    config.channel[0].rx.path = 2; // TODO: replace with string names
+    config.channel[0].rx.path = rxPath;
     config.channel[0].rx.calibrate = false;
     config.channel[0].rx.testSignal.enabled = false;
 
     config.channel[0].tx.enabled = false;
     config.channel[0].tx.sampleRate = sampleRate;
     config.channel[0].tx.oversample = 2;
-    config.channel[0].tx.path = 2; // TODO: replace with string names
+    config.channel[0].tx.path = rxPath;
```
rjonaitis:
```suggestion
    config.channel[0].tx.path = 0;
```
in general case assigning index of rxPath value to Tx path might not always be valid. Since this is just Rx example, Tx path can be set to 0.

rjonaitis (changes requested):
looks good.

PR #103, opened 2024-10-19
Fix Wunused-function build warning

Gluttton:
See the commit message for details.


rjonaitis:
That function is part of work in progress refactoring of registers handling. So let's keep it.

PR #111, opened 2025-01-15
LimeSDR: Add ADF4002 to limeGUI

lifeisafractal:
Add the ADF4002 to the LimeSDR's socTree such that it will show up in limeGUI.

This was tested on a LimeSDR board and using the GUI interface was confirmed to properly affect the chip. Muxout was changed to both DVdd and DGND and confirmed at R160 with a multi-meter.


rjonaitis (approved):
looks good to me, thanks.

PR #116, opened 2025-02-12
support fedora dependencies

AnandSingh:
have tested on fedora 41

### DIFF
--- a/install_dependencies.sh
+++ b/install_dependencies.sh
@@ -14,6 +14,8 @@ TYPE=""
 
 if [[ $ID == "ubuntu" || $ID == "debian" ]]; then
   TYPE=$ID
+elif [[ $ID == "fedora" ]]; then
+  TYPE="fedora"
 fi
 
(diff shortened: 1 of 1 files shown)


install_dependencies.sh
```diff
@@ -46,18 +48,32 @@ if [[ $TYPE == "ubuntu" ]]; then
...
-  build-essential \
-  cmake \
-  $headers \
-  libglew-dev \
-  libsoapysdr-dev \
-  libusb-1.0-0-dev \
-  $libwx
+if [[ $TYPE == "fedora" ]]; then
+  dnf install $yes \
+    @development-tools \
```
rjonaitis, line 63: what for is the '@' symbol?

AnandSingh: this will install the development-tools group which is equivalent to build-essential in ubuntu, also alternatively the groups can be install in fedora as below

```
sudo dnf group install development-tools
```

PR #132, opened 2025-05-04
Enable cross compile with Mingw and MXE

oe1rsa:
The mingw compiler with a linux machine as the build host for windows target needs a few changes to cmake files and several source files.

For cross building I was using the MXE toolset (https://mxe.cc). This toolset provides a cmake executable together with a wrapper script, that includes an appropriate platform file for cmake.

From the available compilers I choose GCC 13 and from the libraries I used libusb1, libftdi1, wxwidgets, and glew.

Since windows compilers and linux gcc differ in what long and long long means for 64 bit types I used the printf macros from the #include \<cinttypes\> to address several warnings (btw. these warnings also appear when using MSVC).

I have not managed to do a complete test, but I was able to launch the GUI and connect to a lime Mini v2. Since I have no HW windows machine I tested this from a QEMU based virtual machine.

Addressed issues from test runs: 32 bit windows failing and clang-format requirements.


rjonaitis:
Looks ok. I'm not familiar with the MXE, or how Windows handles libusb. The LimeSDR Mini v2 might work with this cross compilation with the libusb1, libftdi1 libraries linked, but the LimeSDR-USB most likely won't as it uses Cypress FX3 usb controller which has a different driver on Windows.

Issue #131, opened 2025-04-21
find_package(limesuiteng REQUIRED) not finding kissfft


oe1rsa:
After a src build and local install, in another project find_package(limesuiteng REQUIRED) will not find kissfft library.

Conditions tested: Built limesuiteng with static libraries.

oe1rsa:
I found a work-around: One has to install kissfft separately and the insert in the project:

    find_package(Threads REQUIRED)
    find_package(kissfft 131.1.0 REQUIRED)
    find_package(limesuiteng 0.3.0 REQUIRED)

for this to work, however, another problem needs to be addressed, for which I create a separate issue and pull request.

I think the correct way would be to use a find_dependency call in the limssuitengConfig.cmake package.

rjonaitis:
The static libraries issue is caused by the project compiling kissfft from source into it's own static library (kissfft.a), and then linking it into limesuiteng.
When the limesuiteng is built as a shared library, it absorbs the kissfft.a code and there is not external dependency left.

When the limesuiteng is built as static library, the kissfft.a becomes a dependency. The limesuiteng install step does not install the produced kissfft.a into system directory, thus external project that link to static version of limesuiteng will not find it, unless it is already installed on the system. As the kissfft.a is built from source LimeSuiteNG does not install kissfft.a into system, to not overwrite the library that might already be installed from package manager as the versions might be different.

The solution would be to add kissfft sources directly to the limesuiteng target, therefore embedding the code into limesuiteng.a library. As the kissfft is also used in GUI, that would result in compiling kissfft code twice, unless it would be compiled into cmake target objects.

oe1rsa:
You wrote:

```
As the kissfft.a is built from source LimeSuiteNG does not install kissfft.a into system, to not overwrite the library that might already be installed from package manager as the versions might be different.
```
The problem I see, with this approach, that one likely will get symbol clashes when trying to link also against an external (possibly system installed) library. Wouldn't it be better to depend on its object files instead and strip them after linking? Similar to how lms_gfir is linked against.

rjonaitis:
So far the main use case of limesuiteng is as shared library, so the use cases of it being static library has not been thoroughly tested.
Also the package manager does not seems to install static version of kissfft, only a shared library.

> Similar to how lms_gfir is linked against.

Yes, that's what i meant by solution of making kissfft a objects library.

oe1rsa:
[...]

PR #137, opened 2025-06-14
for static build export kissfft as a private lib

oe1rsa:
When statically linking to liblimesuiteng one
also needs to link against the kissft lib.
Since the user may or may not have installed
kissfft as a static lib, we provide a private
version which is linked as liblimesuiteng-kissfft-float.a when linking against liblimesuiteng.
When the user also has installed the upstream
kissfft she may link against this version, but
then the results become dependent on where the
both libraries appear on the linker line.
Usually the first wins. So take care!

This pull request will resolve issue #131

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -98,6 +98,19 @@ install(
     RUNTIME DESTINATION bin # .dll file
 )
 
+if(NOT BUILD_SHARED_LIBS)
+    # install with a prefix, so will not interfere with kissfft lib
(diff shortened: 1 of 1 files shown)


rjonaitis:
dependency on kissfft.a has been removed by integrating it's source into limesuiteng.a b742fc67d980374a599e0882e8997925872dad73

oe1rsa:
Thank you! Works for me. 
Btw.: I like your solution, you took the harder route, however.

PR #144, opened 2025-07-29
LimeSuite & GNURadio plugin documentation update

Arnius:
List of updates:

1. Lime Suite NG build documentation for Windows conda environment (Using CMake or Conda-Build tools).
2. GNURadio-limesuiteng plugin build documentation for Windows conda environment (Using CMake or Conda-Build tools).
3. Added conda recipes (local recipe and for CI/CD systems) for Lime Suite NG and gnuradio-limesuiteng plugin.
4. Updated gnuradio-limesuiteng python bindings automatic hash regeneration.


docs/gettingStarted/plugins/gnuradio/Windows/buildWindows.rst
```diff
@@ -0,0 +1,96 @@
...
+   conda activate -n <environment name>
+
+.. hint::
+   Check out radioconda and conda environment set up process. See :ref:`radioconda-setup-ref`.
+
+Install conda packages:
+
+.. code-block:: bash
+
+   conda install conda-build conda-forge-pinning gnuradio=3.10.9.2 gnuradio-build-deps=3.10.9.2 boost=1.82 vs2022_win-64 cmake=3.26.4 python=3.12.9 pybind11=2.11.1 mpir=3.0.0
```
rjonaitis: To avoid discrepancies in the documentation and other places where these packages are listed it would be best to put this data into a file, and then pass it to the conda intall command.

docs/gettingStarted/plugins/gnuradio/Windows/buildWindows.rst
```diff
@@ -0,0 +1,96 @@
...
+   #. MSVC v143 - VS 2022
+   #. Windows 11 SDK
+
+#. Conda packages:
+
+   #. conda-build
+   #. conda-forge-pinning
+   #. gnuradio=3.10.9.2
+   #. gnuradio-build-deps=3.10.9.2
+   #. boost=1.82
```
rjonaitis, line 22: Not all of these packages are direct dependencies for gr-limesuiteng. I assume conda would automatically install 'boost, mpir' packages as part of gnuradio package installation, so you don't need to list them explicitly.

docs/gettingStarted/plugins/gnuradio/index.rst
```diff
@@ -0,0 +1,50 @@
...
+   Found 1 device(s) :
+   0: LimeSDR Mini, addr=0403:601f, serial=00000000000000
+
+Launch gnuradio in conda environment (or in linux terminal) by executing the following command:
+
+.. code-block:: bash
+
+   (LimeSuiteNG) D:\LimeSuiteNG\plugins\gr-limesuiteng>gnuradio-companion
+
+Open ``FM_receiver.grc`` example from repository ``<repo root>\plugin\gr-limesuiteng\examples`` directory. Enter SDR serial number as shown in figure below and run the example.
```
rjonaitis: You can directly provide the filename to be opened for the gnuradio-companion command
`gnuradio-companion somefile.grc`

There is no need to enter the serial number, if host system has only one LimeSDR plugged in, then it will automatically be used.
Specifying device handle is only necessary to select from multiple devices.

docs/development/buildSource/buildWindows.rst
```diff
@@ -6,37 +9,82 @@ Prerequisites
...
+
+.. hint::
+   Check out radioconda and conda environment set up process. See :ref:`radioconda-setup-ref`.
+.. Add reference to radioconda and conda setup
+
+Install conda packages:
+
+.. code-block:: bash
+
+   conda install conda-build conda-forge-pinning vs2022_win-64 cmake=3.26.4 ninja
```
rjonaitis: why a specific version of cmake is needed?

docs/gettingStarted/plugins/gnuradio/Linux/buildLinux.rst
```diff
@@ -0,0 +1,62 @@
...
+
+Required components to compile gnuradio-limesuiteng plugin:
+
+#. GNURadio (version 3.10.9.2)
+#. Boost    (version 1.83 or higher)
+#. Pybind11 (version 2.11.1 or higher)
+#. Gmp
+#. Python   (version 3.12.3 or higher)
+#. C++ compiler (GCC major version 13.xx)
+#. CMake (version 3.28.3 or higher)
```
rjonaitis: There is no need for specific cmake version.

docs/gettingStarted/plugins/gnuradio/Linux/buildLinux.rst
```diff
@@ -0,0 +1,62 @@
...
+#. LimeSuiteNG
+
+.. Note::
+   Some plugin component versions may be incompatible and can break the functionality of a plugin. For example, higher versions of pybind11 can cause functionality problems as the GNURadio itself uses 2.11.1 version of pybind11. It is recommended to install the specified minimum version of component from required component list.
```
rjonaitis: Assuming the dependencies are installed using linux package manager, then there will be no versions incompatibility.

docs/gettingStarted/plugins/gnuradio/Linux/buildLinux.rst
```diff
@@ -0,0 +1,62 @@
+Build from source
+=================
+
+Prerequisites
+-------------
+
+Required components to compile gnuradio-limesuiteng plugin:
+
+#. GNURadio (version 3.10.9.2)
```
rjonaitis: On Linux the plugin is compatible with all GNURadio versions >=3.10

docs/gettingStarted/plugins/gnuradio/Windows/condaRecipe.rst
```diff
@@ -0,0 +1,34 @@
...
+
+.. important::
+
+   Compared to standard build procedure, building plugin package using conda-build tool takes more time (5-8 min.) and resources. For development purposes it is advised to build from source using cmake. See :ref:`windows-plugin-build-ref`.
+
+.. hint::
+   
+   Check out radioconda and conda environment setup. See :ref:`radioconda-setup-ref`.
+
+Gnuradio-limesuiteng plugin for GNURadio 3.10.9.2 (Python 3.12) can also be built using conda recipe. Plugin recipe files are provided in ``<project>\plugins\gr-limesuiteng\.conda\`` directory.
```
rjonaitis: no need to specify exact versions everywhere in general descriptions. Versions info is provided within the build scripts.

docs/development/buildSource/buildWindows.rst
```diff
@@ -6,37 +9,82 @@ Prerequisites
...
+
+   #. MSVC v143 - VS 2022
+   #. Windows 11 SDK
+
+#. Conda packages:
+
+   #. conda-build
+   #. conda-forge-pinning
+   #. vs2022_win-64
+   #. cmake (supported versions by project: 3.15 - 3.31)
```
rjonaitis: should work with any cmake version. The minimal required cmake version check is done by cmake itself, and will produce a error message it the version is too low.

plugins/gr-limesuiteng/.conda/recipe/meta.yaml
```diff
@@ -29,67 +22,54 @@ requirements:
...
+    - numpy =1.26.4
+    - volk =3.1.2
 
   run:
-    - numpy
-    - python
-    # Add/remove runtime dependencies here
+    - numpy =1.26.4
+    - python =3.12.9
+    - gnuradio-core =3.10.9.2
```
rjonaitis, line 48: any gnuradio version should work that's >=3.10

rjonaitis (changes requested):

plugins/gr-limesuiteng/python/limesuiteng/bindings/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ endif(NOT limesuiteng_sources)
 gr_python_check_module_raw("pygccxml" "import pygccxml" PYGCCXML_FOUND)
 
 # GrPybind cmake scripts that were before 3.10.7 had wrong paths when used from nested project
-string(COMPARE LESS ${Gnuradio_VERSION} "3.10.7.0" USE_LOCAL_GRPYBIND)
+string(COMPARE LESS ${Gnuradio_VERSION} "3.10.7.0"  USE_LOCAL_GRPYBIND)
```
rjonaitis: fix formatting

docs/development/buildSource/buildWindows.rst
Arnius: During testing I found that conda installs latest CMake 4.x version which causes build files configuration error with the message that is displayed in the image. In this particular case it doesn't really matter if the version is 3.26, 3.28 or any other, as long as the version stays bellow 4.x. I specified the valid CMake version range in components list (when building LimeSuiteNG for Windows). I believe that the version range specification is necessary, since it helps the users to avoid this problem entirely in the first place. I also tried to add the -DCMAKE_POLICY_VERSION_MINIMUM=3.5 flag, but it had no effect.
image

docs/development/buildSource/buildWindows.rst
Arnius: During testing I found that conda installs latest CMake 4.x version which causes build files configuration error with the message that is displayed in the image. In this particular case it doesn't really matter if the version is 3.26, 3.28 or any other, as long as the version stays bellow 4.x. I specified the valid CMake version range in components list (when building LimeSuiteNG for Windows). I believe that the version range specification is necessary, since it helps the users to avoid this problem entirely in the first place. I also tried to add the -DCMAKE_POLICY_VERSION_MINIMUM=3.5 flag, but it had no effect.
image

rjonaitis: limesuiteng should be updated to work with CMake 4.
To do that, patch cmake minimum required version to 3.5 in external dependency archives:
https://github.com/myriadrf/LimeSuiteNG/blob/develop/external/args.zip
https://github.com/myriadrf/LimeSuiteNG/blob/develop/external/rang.zip

plugins/gr-limesuiteng/python/limesuiteng/bindings/CMakeLists.txt
Arnius: Formatting fixed.

docs/development/buildSource/buildWindows.rst
Arnius: Updated CMake required minimum version parameter to 3.15.

docs/development/buildSource/buildWindows.rst
Arnius: Update CMake minimum required version parameter to 3.15 for args and rang packages.

docs/gettingStarted/plugins/gnuradio/Linux/buildLinux.rst
Arnius: Removed CMake version requirements.

docs/gettingStarted/plugins/gnuradio/index.rst
Arnius: Expanded on device connectivity and example usage in GNURadio plugin demonstration section.

docs/gettingStarted/plugins/gnuradio/Windows/condaRecipe.rst
Arnius: Removed GNURadio version strings from general description.

docs/gettingStarted/plugins/gnuradio/Linux/buildLinux.rst
Arnius: Updated GNURadio version requirements to 3.10.x.x.

docs/gettingStarted/plugins/gnuradio/Linux/buildLinux.rst
Arnius: Removed this note.

docs/gettingStarted/plugins/gnuradio/Windows/buildWindows.rst
Arnius: Removed mpir and gnuradio-build-deps packages as these are in fact not necessary. However, the boost=1.82 package is required because it provides boost dev libraries and headers which do not come alongside the gnuradio package. Without the package cmake config fails:
image

Gnuradio package only provides release boost libraries which are required at run time.

PR #151, opened 2025-09-28
Fix another bogus arg order for log message

oe1rsa:
Another small issue, likely a copy and paste error of the previous reported issue.


oe1rsa:
Sorry for having put this into a second pull request. I have overseen that the issue was present a second time in the file.

oe1rsa:
Ok, sorry I am still struggling with the arcane github interface. Closing was not what I had inteded.

PR #158, opened 2025-11-29
debian: remove obsolete limepcie-dkms.dkms

MrCry0:
There is an actual dkms config file template
(drivers/linux/limepcie/dkms.conf.in).
Remove obsolete static dkms config file (debian/limepcie-dkms.dkms) to fix
limepcie-dkms package install issue [1].

[1]
Setting up limepcie-dkms:amd64 (0.1.10) ...
Loading new limepcie/0.1.9 DKMS files...
...
Building for 6.14.0-35-generic and 6.14.0-36-generic
...
Error! The directory /var/lib/dkms/limepcie/0.1.9/source does not appear to have module source located within it.
Build halted.
...

### DIFF
--- a/debian/limepcie-dkms.dkms
+++ b/debian/limepcie-dkms.dkms
@@ -1,7 +1,0 @@
-PACKAGE_NAME="limepcie"
-PACKAGE_VERSION="0.1.9"
-MAKE[0]="make"
-CLEAN="make clean"
-BUILT_MODULE_NAME[0]="limepcie"
-DEST_MODULE_LOCATION[0]="/updates"
-AUTOINSTALL="yes"


rjonaitis:
The failure was due to mismatch of module versions in the debian config and the actual module config files.
The debian/limepcie-dkms.dkms file has to be present, so that debian packaging tools would detect it and setup post installation and pre removal scripts for the package.
Without this file, the source code and dkms.conf file would be installed to the system directories, but without those scripts, it woudn't be registered into the actual DKMS system.

I've made a fix to ensure the dkms configuration files will match https://github.com/myriadrf/LimeSuiteNG/commit/552c4e7bc6ec216a96be898440d9c8596c35edfd

MrCry0:
> The failure was due to mismatch of module versions in the debian config and the actual module config files. The debian/limepcie-dkms.dkms file has to be present, so that debian packaging tools would detect it and setup post installation and pre removal scripts for the package. Without this file, the source code and dkms.conf file would be installed to the system directories, but without those scripts, it woudn't be registered into the actual DKMS system.
> 
> I've made a fix to ensure the dkms configuration files will match [552c4e7](https://github.com/myriadrf/LimeSuiteNG/commit/552c4e7bc6ec216a96be898440d9c8596c35edfd)

Ah, sure thing, you're right. Thanks for pointing this!

But I don't like solutions that change something in CMAKE_SOURCE_DIR/CMAKE_CURRENT_SOURCE_DIR on a project building.

I'll prepare a solution without such changes.

rjonaitis:
> But I don't like solutions that changes something in CMAKE_SOURCE_DIR/CMAKE_CURRENT_SOURCE_DIR on a project building.

Agreed, but in this case, I see no other option how to ensure debian/ directory data to be up to date with the data from source locations, besides manual editing.

> There is an actual dkms config file template (drivers/linux/limepcie/dkms.conf.in).
The template is for CMake to configure/generate the output dkms.conf file in build directory. The template cannot be directly used by DKMS.
The project is setup to essentially generate a final self sufficient kernel module directory (build/drivers/linux/limepcie/limepcie-0.1.10), so it could be just copied into the system and then build with just the 'make' tools or DKMS.

Replacing debian/limepcie-dkms.dkms file with a symlink to generated config file in the build directory is not reliable as that would require hardcoded build directory path.
Removing the limepcie-dkms.dkms file and explicitly adding the postinstall preremove scripts, has the same problem, they would need to have the module version updated.

I'm not an expert in debian packaging, but it seems that there is no way to make dynamic changes for packaging without modifying the ${CMAKE_SOURCE_DIR}/debian files. There's the debian/rules script, I assume it could generate/override some parameters, but it requires extensive knowledge of the debian packaging tools and process.

> I'll prepare a solution without such changes.

I'll look forward to it.

MrCry0:
@rjonaitis done. please review the https://github.com/myriadrf/LimeSuiteNG/pull/161

PR #161, opened 2025-12-08
limepcie: avoid modifying source files during build

MrCry0:
It is not recommended to modify source files when building a project.

Introduce a drivers/linux/limepcie/VERSION file containing the limepcie driver version, and reuse it for the module and Debian DKMS configuration files to avoid rewriting the debian/limepcie-dkms.dkms file in the source tree.

Fixes:   552c4e7b ("limepcie: Auto update dkms configuration file in debian directory")


rjonaitis (changes requested):
Now that the version in dkms.conf is templated,
the version has to be also explicitly set for "dkms add" command in the cmake install procedure https://github.com/myriadrf/LimeSuiteNG/blob/75b603dfcd17b318e0d7ba1d6497ce3a19650b3a/drivers/linux/limepcie/cmake/add_kernel_module.cmake#L246

MrCry0:
> Now that the version in dkms.conf is templated, the version has to be also explicitly set for "dkms add" command in the > cmake install procedure

@rjonaitis 

What issue does it solve?

Now this command also copies the project to /usr/src/ due to semantic "dkms add /path/to/source-tree". If the command is changed to "dkms add module/version" sematic, copying should be done before this command with extra cmake command.

rjonaitis:
Never mind, I mistaken that the PACKAGE_VERSION="#MODULE_VERSION#" template was also placed in the source directory dkms.conf.
This change affects only the debian dkms.conf , so all good.

PR #165, opened 2025-12-22
LMS64C protocol update

Arnius:
Updates:

1. Updated LMS64C protocol to support CSR register read/write operations.
2. Added CSR module to LimeSuiteNG GUI application.
3. Added limeCSR CLI application for writing/reading LimeSDR CSR registers from command line.
4. Added limeGPSDO CLI application for controlling LimeSDR GPSDO from command line.


cli/limeGPSDO.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+   // Logger members
+   string getGPSDOStatusMsg();
+
+   private:
+   void formatGPSDOStatus(uint64_t state, uint64_t accuracy, uint64_t tpulse, array<string, 3>& GPSDOStatus);
+
+   ICSR * mCSR_interface;
+   std::string mGPSDOStatusMsg;
+   static gpsdo_reg_list_t * mpGPSDORegisterList;
```
rjonaitis: should not be 'static', it would break things if there would be multiple instances of GPSDODriver for different boards.
Also should be 'const' to protect against modifying the list.

cli/limeGPSDO.cpp
```diff
@@ -0,0 +1,606 @@
...
+      if(mHandle.addr == "0403:601f"s || mHandle.addr == "374d:0019"s)
+      {   
+         mpGPSDORegisterList = &SDR_GPSDO_Registers[DeviceID::LIMESDR_MINI_V2];
+         regListUpdated = true;
+      }
+      lime::debug("DEBUG: Selected CSR register list for LimeSDR Mini V2");
+      break;
+   
+   case MediaType::PCIE:
+      if(mHandle.name.find("XTRX"s))
```
rjonaitis: string::find() if pattern is not found returns string::npos which is (-1), so this condition will pass for all devices.

cli/limeGPSDO.cpp
```diff
@@ -0,0 +1,606 @@
...
+         headerLength = printHeader();
+
+      results.dumpCount = ++dumpCount;
+      cout << formatToFit(&results, headerLength) << endl;
+      
+      std::this_thread::sleep_for(delay);
+      continueLoop = (dumpCount < numDumps || numDumps == 0);
+
+      // This is for infinite loop. To prevent UB, reset counter
+      if(dumpCount == std::numeric_limits<int>::max())
```
rjonaitis: make it 'unsigned', unsigned integers have defined overflow behaviour, only signed integers overflow is UB.

cli/limeGPSDO.cpp
```diff
@@ -0,0 +1,606 @@
...
+   else if(state == 0ULL)
+      GPSDOStatus[GPSDO_STATE] = "Coarse tune";
+   else
+      GPSDOStatus[GPSDO_STATE] = "Unknown";
+   
+   if(accuracy < 4)
+      GPSDOStatus[GPSDO_ACCURACY] = GPSDODriver::accuracyLevelList[accuracy];
+   else
+   {
+      GPSDOStatus[GPSDO_ACCURACY] += "Unknown(";
```
rjonaitis: Is it suppose to just append, no need to initialize starting value?

cli/limeGPSDO.cpp
```diff
@@ -0,0 +1,606 @@
...
+   
+   default:
+      mpGPSDORegisterList = nullptr;
+      break;
+   }
+
+   return regListUpdated;
+}
+
+void GPSDODriver::formatGPSDOStatus(uint64_t state, uint64_t accuracy, uint64_t tpulse, array<string, 3>& GPSDOStatus)
```
rjonaitis: Does not need to be a member function, it does not operate on object's data.
This just converts values into strings, might as well be three separate functions for each value instead of passing reference to array of strings, that would be more clear.

cli/limeGPSDO.hpp
```diff
@@ -0,0 +1,185 @@
...
+#include <csignal>
+#include <unordered_map>
+#include <cmath>
+#include <atomic>
+#include <limits>
+#include <cstdio>
+#include <cinttypes>
+#include "args.hxx"
+
+using namespace std;
```
rjonaitis: 'using namespace' in header files is bad practice, because it will leak to all other source files that include it and can cause issues.

cli/limeGPSDO.hpp
```diff
@@ -0,0 +1,185 @@
...
+   PPSDO_STATUS_HUNDRED_S_ERROR = 9,
+   PPSDO_STATUS_DAC_TUNED_VAL = 10,
+   PPSDO_STATUS_ACCURACY = 11,
+   PPSDO_STATUS_PPS_ACTIVE = 12,
+   PPSDO_STATUS_STATE = 13
+};
+
+enum class DeviceID : uint8_t
+{
+   LIMESDR_XTRX = 0,
```
rjonaitis: no need for new enumeration, can use enum from https://github.com/myriadrf/LimeSuiteNG/blob/develop/src/protocols/LMSBoards.h

cli/limeGPSDO.hpp
```diff
@@ -0,0 +1,185 @@
...
+};
+
+enum class MediaType : uint8_t
+{
+   USB = 0,
+   PCIE = 1,
+   UNDEFINED = 2
+};
+
+static unordered_map<DeviceID, gpsdo_reg_list_t> SDR_GPSDO_Registers =
```
rjonaitis: should be const, and initialized in .cpp file. 'static' in a header file would result in multiple copies of the same data for each .cpp that includes it.

rjonaitis (changes requested):

rjonaitis:
For packaging build to not fail, need to update installed cli application list https://github.com/myriadrf/LimeSuiteNG/blob/develop/debian/limesuiteng-cli.install

9600:
Just wondered whether this would also be used with alternate PPS sources, e.g. 1588 PTP Ethernet, and corresponds to the LimePPSDO gateware functionality? If so, should this be "limePPSDO"?

cli/limeGPSDO.cpp
```diff
@@ -52,68 +114,90 @@ uint64_t GPSDODriver::getSigned32bit(uint64_t address, OpStatus * status)
 
 uint64_t GPSDODriver::get_1s_error(OpStatus * status)
 {
-   return this->getSigned32bit((*mpGPSDORegisterList)[GPSDORegistersID::PPSDO_STATUS_ONE_S_ERROR], status);
+   return this->getSigned32bit((*mpGPSDORegisterList).find(GPSDORegistersID::PPSDO_STATUS_ONE_S_ERROR)->second, status);
```
rjonaitis: Seems odd, why do dereferencing`(*mpGPSDORegisterList).`when `mpGPSDORegisterList->` would work.
No need to use `.find()->second`, const unordered map can be accessed with `.at(GPSDORegistersID::PPSDO_STATUS_ONE_S_ERROR)`. Map constness does not allow use of [], as it creates a new element in the container if the key was not found.

rjonaitis (changes requested):

rjonaitis:
Also rename limeGPSDO to limePPSDO

PR #170, opened 2026-01-21
Avoid polling by waiting for DMA complete interrupt

amurray-thegoodpenguin:
Let's reduce CPU utilisation by waiting for an interrupt rather than polling. This reduces CPU usage by approximately 25%.


drivers/linux/la9310_limesdr/la9310_base.h
```diff
@@ -333,6 +303,11 @@ struct la9310_dev {
 
     struct la9310_mem_region_info dmem_proxy;
     struct platform_device* uart;
+
+    u8 arm_m4_fw_loaded;
+    u8 vspa_fw_loaded;
```
rjonaitis: arm_m4_fw_loaded and vspa_fw_loaded flags have been already removed.

src/comms/PCIe/LA9310_PCIe.cpp
```diff
@@ -169,6 +169,17 @@ mmaped_region LA9310_PCIe::GetBar(uint8_t i)
     return mapped_ranges[i];
 }
 
+void LA9310_PCIe::wait_for_new_data()
+{
+    while(1) {
+        if (ioctl(mFileDescriptor, LA9310_IOCTL_WAIT_FOR_DATA) < 0){
+	    if (errno == EINTR)
```
rjonaitis: LA9310_IOCTL_WAIT_FOR_DATA returns value of la9310_wait_for_data(), the available values are just 0 or -ERESTARTSYS.

src/comms/PCIe/LA9310_PCIe.cpp
```diff
@@ -169,6 +169,17 @@ mmaped_region LA9310_PCIe::GetBar(uint8_t i)
     return mapped_ranges[i];
 }
 
+void LA9310_PCIe::wait_for_new_data()
```
rjonaitis: Should have a timeout parameter, so that just in case of device malfunction the software would not get stuck in infinite loop.

rjonaitis (changes requested):
Fix source formatting.
You can ignore the build error on MSVC builder, it's not related to this PR.

drivers/linux/la9310_limesdr/la9310_base.h
amurray-thegoodpenguin: Ah good spot! These must have crept in when rebasing to the latest. I'll remove.

rjonaitis:
I've merged the relavant firmware updates into their repositories.
You should build and add updated firmware files into this PR:
LimeSuiteNG/firmware/limesdr-micro/la9310.bin
LimeSuiteNG/firmware/limesdr-micro/apm-iqplayer.bin (1T1R version)

src/comms/PCIe/LA9310_PCIe.cpp
amurray-thegoodpenguin: That's true, but the value can be changed by the signal handler in the kernel, e.g. arch/arm64/kernel/signal.c

rjonaitis: Not sure how the signal handler could do that, you are not forwarding kernel's return value from (wait_for_completion_interruptible), but just returning a hardcoded const value (return -ERESTARTSYS) https://github.com/myriadrf/LimeSuiteNG/pull/170/files#diff-28187cf9a3f645769376c65869618b337138083d3375d7fd05ddd4c75462513eR40

amurray-thegoodpenguin: I was referring to this  https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/kernel/signal.c#n1630

rjonaitis: ok, so the 'ioctl' call itself can be interrupted and return that code? Brief look into 'ioctl' man does not mention EINTR as one of it's return codes (although kernel doc mentions: ioctl commands can return negative error codes as documented in errno(3);). 
A comment should be added about this, otherwise it's confusing why would this code be returned, without diving into the kernel source.

PR #174, opened 2026-02-10
gr-limesuiteng: expose linkFormat, always output fc32

spetca:
gr-limesuiteng: linkFormat and gr output issues

The gr-limesuiteng GNURadio blocks (sdrdevice_source / sdrdevice_sink) previously had two bugs and one missing feature:

 ## Bug 1 — linkFormat was hardcoded to I16
sdrdevice_block_base.cc unconditionally set streamCfg.linkFormat = DataFormat::I16 regardless of what the user configured. This meant the XTRX (and other LimeSuiteNG devices) always used 4 bytes/sample on the PCIe DMA bus, making 120 MSPS SISO impossible on a PCIe Gen2 x1 link (~400 MB/s limit: I16 needs 480 MB/s, I12 needs only 360 MB/s).

 ## Bug 2 — dataFormat parameter created type mismatch
The sc16 option advertised a 4-byte output port but work() unconditionally wrote 8 bytes (F32) into it — a silent buffer overflow causing garbage output and GNURadio port type mismatch errors when connecting to downstream fc32 blocks.

 ## Fix
  - streamCfg.format is now hardcoded to F32. The GNURadio ports are always fc32. TRXLooper handles the integer→float
  conversion internally before Receive() returns — this is a CPU SIMD operation, not a bottleneck.
  - linkFormat is now a user-configurable parameter (complex16_t = I16 default, complex12_t = I12 for PCIe bandwidth
  saving). Setting I12 writes the FPGA register 0x0008 to pack samples as 3 bytes/IQ pair and programs the DMA
  transfer size accordingly — real hardware behavior, not a software conversion.
  - The old dataFormat parameter is removed entirely. The link_format GRC block parameter replaces it.

## changelog 

  - plugins/gr-limesuiteng/
      - lib/sdrdevice_block_base.h/.cc       — removed dataFormat, added linkFormat, hardcoded format=F32
      - lib/sdrdevice_source_impl.h/.cc      — removed dataFormat, io_signature always sizeof(complex32f_t)
      - lib/sdrdevice_sink_impl.h/.cc        — same
      - include/gnuradio/limesuiteng/
      - sdrdevice_source.h                 — updated make() signature, added docs
      - sdrdevice_sink.h                   — same
   - python/limesuiteng/bindings/
      - sdrdevice_source_python.cc         — updated pybind11 args
      - sdrdevice_sink_python.cc           — same
   - grc/
      - limesuiteng_sdrdevice_source.block.yml  — removed data_format, output hardcoded fc32
      - limesuiteng_sdrdevice_sink.block.yml    — same

### DIFF
--- a/plugins/gr-limesuiteng/grc/limesuiteng_sdrdevice_sink.block.yml
+++ b/plugins/gr-limesuiteng/grc/limesuiteng_sdrdevice_sink.block.yml
@@ -6,6 +6,17 @@ flags: throttle
 file_format: 1
 
 parameters:
+- id: data_format
+  label: Data Format
+  dtype: enum
+  options: [fc32, sc16]
+  option_labels: ["Complex Float32", "Complex Int16"]
+  option_attributes:
+    limeFormat: ["complex32f_t", "complex16_t"]
+  default: fc32
+  hide: all
+  category: General
+
 - id: link_format
   label: Link Format
   dtype: enum
@@ -128,7 +139,7 @@ asserts:
 templates:
   imports: from gnuradio import limesuiteng
   make: |-
-    limesuiteng.sdrdevice_sink(${alias}, ${deviceHandleHint}, ${chip_index}, ${ channel_indexes }, "${ link_format.limeFormat }", ${sample_rate}, ${rf_oversampling.value})
+    limesuiteng.sdrdevice_sink(${alias}, ${deviceHandleHint}, ${chip_index}, ${ channel_indexes }, "${ data_format.limeFormat }", "${ link_format.limeFormat }", ${sample_rate}, ${rf_oversampling.value})
     self.${id}.set_lo_frequency(${rf_freq})
     self.${id}.set_lpf_bandwidth(${lpf})
     self.${id}.set_gfir_bandwidth(${gfir_bandwidth})
--- a/plugins/gr-limesuiteng/grc/limesuiteng_sdrdevice_sink.block.yml
+++ b/plugins/gr-limesuiteng/grc/limesuiteng_sdrdevice_sink.block.yml
@@ -12,19 +12,20 @@ parameters:
   options: [fc32, sc16]
   option_labels: ["Complex Float32", "Complex Int16"]
   option_attributes:
+    type: [fc32, sc16]
     limeFormat: ["complex32f_t", "complex16_t"]
   default: fc32
-  hide: all
+  hide: part
   category: General
 
 - id: link_format
   label: Link Format
   dtype: enum
   options: [i16, i12]
-  option_labels: ["I16 - 4 bytes/sample (default)", "I12 - 3 bytes/sample (XTRX bandwidth saving)"]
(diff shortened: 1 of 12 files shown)


rjonaitis:
The linkFormat addition is welcome, but it shouldn't remove the dataFormat option, you can just set it to be hidden from the block properties.
Indeed GRC does not utilize the sc16 format, but it is available in GNURadio itself, so it should be available if someone wants to use it through code.
As for the memory overflow, you can just add a switch in the work function to cast into different pointer types based on data format.

spetca:
Thanks for the feedback @rjonaitis i believe i captured the recommendations in this new commit

plugins/gr-limesuiteng/grc/limesuiteng_sdrdevice_sink.block.yml
```diff
@@ -7,14 +7,25 @@ file_format: 1
...
-  hide: part
+  default: fc32
+  hide: all
+  category: General
+
+- id: link_format
+  label: Link Format
+  dtype: enum
+  options: [i16, i12]
+  option_labels: ["I16 - 4 bytes/sample (default)", "I12 - 3 bytes/sample (XTRX bandwidth saving)"]
```
rjonaitis: Link format is generic parameter, it shouldn't mention specific device. Same for source block.

plugins/gr-limesuiteng/lib/sdrdevice_sink_impl.cc
```diff
@@ -64,21 +54,25 @@ sdrdevice_sink_impl::sdrdevice_sink_impl(const std::string& alias,
...
+                                         const std::string& linkFormat,
                                          double sampleRate,
                                          int rf_oversampling)
     : gr::sync_block((alias.empty()
                           ? fmt::format("sdrdevice_sink[{:s}]", deviceHandleHint.c_str())
                           : alias),
                      gr::io_signature::make(1 /* min outputs */,
                                             channelIndexes.size() /*max outputs */,
-                                            GetDataFormatTypeSize(dataFormat)),
+                                            (dataFormat == "complex16_t")
```
rjonaitis: For a better readability and maintenance, the format string conversion to size would be better done with a dedicated function, like the removed 'GetDataFormatTypeSize(const std::string& formatname)', instead of inline conditionals.

plugins/gr-limesuiteng/grc/limesuiteng_sdrdevice_sink.block.yml
```diff
@@ -7,14 +7,25 @@ file_format: 1
...
+  option_attributes:
+    limeFormat: ["complex16_t", "complex12_t"]
+  default: i16
```
rjonaitis: You can set the default to I12

rjonaitis (changes requested):

spetca:
Thanks for your guidance on this @rjonaitis appreciate the help

PR #178, opened 2026-03-17
Part 2: Update of doxygen/sphinx documentation for LimeSuiteNG library public API.

Arnius:
Contents of the update:

1. Added a new topic - SDR device stream configuration.
2. Added a new topic with examples - SDR device streaming.
3. Added a new topic with examples - library logging functions.
4. Added missing SDR device configuration sub-topics.
5. Implemented formatting fixes.
6. Implemented channel selection for API functions used to set and get test signal generators for SDR devices.
7. Implemented channel selection for API functions used to set and get dc offset correctors for SDR devices.


Arnius:
**A note for the reviewers**

Will also provide an update to the sphinx documentation topic - "For Developers -> Documentation generation guide". Updated docs will include updated documentation build steps for linux and windows.

9600:
@Arnius Thanks. I'm just on with something else and hope to take a look later this week. I'll need to review changes to conf.py also and consider how we can merge these into the standard conf.py used across all projects. The idea is that this never gets modified and all project-specific config lives in project.py. That way when updating Sphinx we can safely just overwrite conf.py and requirements.txt in every project, thereby avoiding the hassle of these being potentially different for each project.

rjonaitis (approved):

Arnius:
Updated scripts for documentation generation both on Linux and Windows platforms. The developer page for documentation generation has also been updated.

rjonaitis (approved):

PR #186, opened 2026-07-09
1.0 API surface + public C API (freeze proposal phases 1-4)

dmitrymmm:
Implements the LimeSuiteNG 1.0 freeze proposal (patches 0001-0013 from the review): the C++ surface resolutions and the first four phases of the C-ABI transition plan.

## C++ surface (review resolutions, patches 0001-0008)

- Deprecate legacy streaming paths (`RFStream::StreamRx/StreamTx`, `SDRDevice::StreamRx` - the previously unmarked outlier) and the C-string `registerLogHandler` overload
- Add `ChannelId` value type + non-virtual delegating overloads on `SDRDevice` (additive, no vtable change); `module` documented as transition-only per the Path B submodule split
- `OpStatus` is `[[nodiscard]]` at the type level - every dropped status return is now a compiler warning
- `StreamConfig`: `std::function` `onStatus` callback appended; RealtimePolicy kept out of the public surface per review decision (RT stays best-effort and internal)
- Antenna addressed **by name** on the 1.0 surface (`SetAntenna(ChannelId, string)` / `GetAntennaName`), resolving via `RFSOCDescriptor::pathNames`; index-based ChannelId overloads removed
- SOVERSION stays `MAJOR.MINOR-ABI` (review decision); defensive returns in `ChannelConfig::GetDirection`
- `RFStream::Receive/Transmit` regain the **timeout parameter** (trailing, defaulted - source-compatible); threaded through TRXLooper / StreamComposite / RFStream_X8, which previously hardcoded inconsistent values. SoapySDR's `readStream/writeStream` need it, and the C API maps `timeout_ms` onto it.

## Public C API (transition plan phases 1-4, patches 0009-0013)

New public header `limesuiteng/c/limesuiteng.h` (C99-clean, compiles as C and C++) + thin `extern "C"` wrapper built into liblimesuiteng:

- **Enumeration / device tree**: `lime_enumerate`, `lime_device_open/close`, `lime_device_as_sdr` (transitional: collapses onto SDRDevice)
- **SDRDevice**: `set_frequency`, `enable_channel`, `set_antenna` (by name), `get_descriptor`
- **Streaming (the hot path)**: `lime_stream_create/start/stop/recv/send/destroy` over `StreamCreate` + `RFStream`; transparent append-only `lime_StreamConfig` leading with a validated `struct_size`; unified `lime_StreamMeta` (flush Tx-only)
- **Subinterface providers**: GPIO get/set, SPI transact (`NULL` when capability absent)
- **Descriptor accessors**: name/serial/rfsoc/channel-count/antenna names - antenna-by-name is discoverable from C
- **Version**: `lime_get_library/api/abi_version`

Fix included: `LIME_C_BUILD` was initially nested inside the `ENABLE_NEW_GAIN_BEHAVIOUR` block, leaving every `lime_*` symbol marked dllimport and unexported on Windows; now unconditional.

## Verification

- Full build zero errors (MinGW GCC 16, Windows; core + CLI + tests)
- Header compiles standalone as C99 (`-Werror`) and C++17
- 54/54 unit tests pass, including new hardware-free suites for `ChannelId` delegation and antenna name<->index resolution (`tests/api/`)
- 32 `lime_*` symbols exported; C99 smoke test links against the DLL, runs, and passes all null-guard/layout checks
- Also includes PPA packaging tooling (`tools/ppa/`), debian/lintian updates, and CI workflow changes

## Not in this PR (later phases)

Whole-state `lime_SDRConfig`/`configure` (phase 4 remainder), stream stats surface, GNSS/memory providers, GUI/CLI migration onto C (phase 5), C++ export deprecation and freeze gates (phases 6-7). The ~700 `[[nodiscard]]` call-site warnings surfaced by the OpStatus change are left for a separate cleanup pass.


debian/rules
```diff
@@ -15,7 +15,7 @@ VERSION_MINOR := $(word 2,$(subst ., ,$(DEB_VERSION)))
 RELEASE_NUM = $(VERSION_MAJOR).$(VERSION_MINOR)
 
 %:
-	dh $@ --with dkms --buildsystem=cmake --parallel --builddirectory=build
```
rjonaitis, line 18: Why remove 'parallel' option?

.gitattributes
rjonaitis, line 1: why force specific EOLs?

qodana.yaml
```diff
@@ -0,0 +1,49 @@
...
+#  severityThresholds:
+#    any: 15
+#    critical: 5
+#  testCoverageThresholds:
+#    fresh: 70
+#    total: 50
+  
+#Qodana supports other languages, for example, Python, JavaScript, TypeScript, Go, C#, PHP
+#For all supported languages see https://www.jetbrains.com/help/qodana/linters.html
+linter: jetbrains/qodana-jvm-community:2026.1
```
rjonaitis: LimeSuiteNG is a C++ project, not Java.

qodana.yaml
rjonaitis, line 1: From a brief overview this Qodana seems pointless for this project. There is no intention to subscribe to it's services, and the free "community" option does not offer much usefulness in terms of C/C++ support.

src/api_c/capi.cpp
rjonaitis, line 1: Even though it's just a thin wrapper, split implementations into separate files based on categories. Just like there's a one C++ class in a file, there should be a matching C wrapper file for that class.

put all the files in a LimeSuiteNG/src/lib directory.

src/api_c/capi.cpp
```diff
@@ -0,0 +1,329 @@
...
+    return static_cast<lime_OpStatus>(sdr(dev)->EnableChannel(module, dir(d), channel, enable));
+}
+
+lime_OpStatus lime_sdrdevice_set_antenna(
+    lime_SDRDevice* dev, uint8_t module, lime_TRXDir d, uint8_t channel, const char* name)
+{
+    if (dev == nullptr || name == nullptr)
+        return lime_OpStatus_InvalidValue;
+    const lime::ChannelId c{ module, dir(d), channel };
+    return static_cast<lime_OpStatus>(sdr(dev)->SetAntenna(c, std::string(name)));
```
rjonaitis: std::string(name) creates copies of the c string. Use std::string_view for forwarding c strings.

src/include/limesuiteng/c/limesuiteng.h
rjonaitis, line 1: no need for "c" subdirectory, put the public header files in src/include/limesuiteng.
Split into multiple header files by categories "classes".

src/include/limesuiteng/c/limesuiteng.h
```diff
@@ -0,0 +1,179 @@
+/*
+ * LimeSuiteNG public C API.
+ *
+ * Stable C ABI over the C++ internal implementation. Opaque handles,
```
rjonaitis: Too much verbosity. This is public C header, it should not contain any information about the internal implementation details.

src/include/limesuiteng/c/limesuiteng.h
```diff
@@ -0,0 +1,179 @@
...
+ * lime_ snake_case entry points, lime_OpStatus return codes.
+ *
+ * This header is the 1.0 public surface. The C++ headers under
+ * limesuiteng/ are the internal implementation, exported only during the
```
rjonaitis: again, too much verbosity. Why is public header documenting internal headers that it should not even be aware of.

src/include/limesuiteng/c/limesuiteng.h
```diff
@@ -0,0 +1,179 @@
...
+typedef struct lime_GPIO          lime_GPIO;
+typedef struct lime_Stream        lime_Stream;
+typedef struct lime_SDRDescriptor lime_SDRDescriptor; /* opaque; read via lime_descriptor_* accessors */
+
+/* ------------------------------------------------------------------ *
+ * Enumeration & device tree.                                         *
+ * ------------------------------------------------------------------ */
+typedef struct { char str[256]; } lime_DeviceHandle;
+
+/* Fills up to 'max' handles; returns the count found (>=0),
```
rjonaitis: Use proper Doxygen compatible comments  syntax, so the API documentation could be generated from them.

src/include/limesuiteng/c/limesuiteng.h
```diff
@@ -0,0 +1,179 @@
...
+    const uint8_t* rx_channels; /* channel indices to stream; may be NULL when the count is 0 */
+    size_t rx_count;
+    const uint8_t* tx_channels;
+    size_t tx_count;
+    lime_DataFormat format; /* samples layout for _recv / _send */
+    lime_DataFormat link_format; /* wire layout Host<->FPGA */
+    double hint_sample_rate_hz; /* 0 = decide internally */
+} lime_StreamConfig;
+
+/* Unified at the C boundary: flush is Tx-only and ignored on Rx. */
```
rjonaitis: There has to be separate meta types for Rx and Tx to avoid confusion which fields are for which direction.

src/include/limesuiteng/c/limesuiteng.h
```diff
@@ -0,0 +1,179 @@
...
+LIME_C_API lime_Kind       lime_device_kind(lime_device* dev);
+LIME_C_API size_t          lime_device_child_count(lime_device* dev);
+LIME_C_API lime_device*    lime_device_child(lime_device* dev, size_t index);
+LIME_C_API lime_SDRDevice* lime_device_as_sdr(lime_device* dev); /* NULL if not an SDR */
+
+/* ------------------------------------------------------------------ *
+ * SDRDevice. Channels are addressed by (module, dir, channel).       *
+ * ------------------------------------------------------------------ */
+LIME_C_API lime_OpStatus lime_sdrdevice_set_frequency(
+    lime_SDRDevice* dev, uint8_t module, lime_TRXDir dir, uint8_t channel, double hz);
```
rjonaitis: don't over constraint parameters to smalest size like uint8_t in public ABI, use either 32bit or size_t, to avoid future necessity to break ABI if the value range requirement would become larger in future.

src/include/limesuiteng/SDRDevice.hpp
rjonaitis, line 1: What's the point of these additional functions? They are not used anywhere besides their own tests.

qodana.yaml
dmitrymmm: Agreed - removed entirely in 6de740f6 (qodana.yaml and the qodana_code_quality.yml workflow); no other references remain in the tree.

qodana.yaml
dmitrymmm: Removed together with the rest of the Qodana config in 6de740f6.

src/include/limesuiteng/c/limesuiteng.h
dmitrymmm: Fixed in 6de740f6: module/channel are now uint32_t across the C surface (entry points, lime_StreamConfig channel lists, lime_descriptor_channel_count). Since the internal C++ addressing is still 8-bit, out-of-range values return InvalidValue instead of being silently truncated - covered by a smoke-test check.

src/include/limesuiteng/c/limesuiteng.h
dmitrymmm: Fixed in 6de740f6: split into lime_StreamRxMeta (filled by _recv) and lime_StreamTxMeta (read by _send), mirroring the internal StreamRxMeta/StreamTxMeta. flush and the timestamp-wait semantics now exist only on the Tx type.

src/include/limesuiteng/c/limesuiteng.h
dmitrymmm: Done in e73b58d4: the c/ subdirectory is gone; headers now live directly in src/include/limesuiteng/, split per class (types_c.h, device_registry_c.h, sdrdevice_c.h, sdrdescriptor_c.h, rfstream_c.h, versioninfo_c.h) with limesuiteng.h as the umbrella (pairing the C++ limesuiteng.hpp). Each compiles standalone as C99 -Werror. They install via the existing public-header glob.

src/api_c/capi.cpp
dmitrymmm: Done in e73b58d4: capi.cpp is split into one C wrapper file per C++ class under src/lib/ (device_registry_c.cpp, sdrdevice_c.cpp, sdrdescriptor_c.cpp, rfstream_c.cpp, versioninfo_c.cpp), shared internals in a non-installed capi_private.h.

src/include/limesuiteng/c/limesuiteng.h
dmitrymmm: Done in e73b58d4: all C API declarations are now Doxygen (@brief/@param/@return, ///< on struct/enum members), so docs can be generated from the headers.

src/include/limesuiteng/c/limesuiteng.h
dmitrymmm: Fixed in e73b58d4: the public headers no longer mention the internal implementation - file-top comments are down to a one-line @brief.

src/include/limesuiteng/c/limesuiteng.h
dmitrymmm: Fixed in e73b58d4 together with the comment above - the transition-window prose is gone from the public headers.

src/include/limesuiteng/SDRDevice.hpp
dmitrymmm: Fair point - trimmed in e73b58d4. The 26 delegating overloads with no callers are removed. Kept: the ChannelId type plus SetAntenna(ChannelId, const std::string&) and GetAntennaName(ChannelId), which implement the antenna-by-name decision and are what the C API's lime_sdrdevice_set_antenna() is built on. If internal call sites later migrate to ChannelId addressing, the overloads can return piecemeal - they are non-virtual and purely additive.

src/api_c/capi.cpp
dmitrymmm: Fixed in 1589444e: SDRDevice::SetAntenna(ChannelId, ...) now takes std::string_view end-to-end, and the C wrapper forwards the caller's string without constructing a temporary std::string.

debian/rules
dmitrymmm: debian/control declares debhelper-compat (= 13), and --parallel has been the default since compat level 10 - the flag was redundant, so dropping it does not change the build (packages still build in parallel). Happy to restore it if you prefer it explicit.

.gitattributes
dmitrymmm: It is scoped, not repo-wide: only /debian/** and tools/ppa/upload.sh are forced to LF. Those files break when checked out with CRLF on Windows (dpkg-source rejects CRLF control files, bash chokes on CRLF scripts), and part of this work was done from a Windows checkout. Repo-wide line endings stay untouched. Can drop the file if you'd rather leave this to developers' core.autocrlf settings.

src/CMakeLists.txt
```diff
@@ -80,6 +85,11 @@ configure_file(${CMAKE_CURRENT_SOURCE_DIR}/SystemResources.in.cpp ${CMAKE_CURREN
 target_sources(
     limesuiteng
     PRIVATE CommonFunctions.cpp
+            lib/device_registry_c.cpp
```
rjonaitis: Manage /lib sources with CMakeLists.txt file within it.

src/include/limesuiteng/sdrdevice_c.h
```diff
@@ -0,0 +1,117 @@
...
+LIME_C_API lime_GPIO* lime_sdrdevice_get_gpio(lime_SDRDevice* dev);
+
+/**
+ * @brief Sets the state of a single GPIO pin.
+ * @param gpio The GPIO subinterface.
+ * @param pin The pin index.
+ * @param value The pin state to set.
+ * @return The status of the operation.
+ */
+LIME_C_API lime_OpStatus lime_gpio_set_value(lime_GPIO* gpio, uint32_t pin, bool value);
```
rjonaitis: lime_sdrdevice just returns the lime_GPIO handle. Everything else regarding lime_GPIO functionality should be in it's own header.

src/include/limesuiteng/sdrdevice_c.h
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * @brief Performs a full-duplex SPI transaction on an internal bus.
+ * @param spi The SPI subinterface.
+ * @param bus_address The chip-select id of the internal bus.
+ * @param mosi Data to write, or NULL for a read-only transfer.
+ * @param[out] miso Buffer for data read, or NULL for a write-only transfer.
+ * @param count The number of 32-bit words to transfer.
+ * @return The status of the operation.
+ */
+LIME_C_API lime_OpStatus lime_spi_transact(
```
rjonaitis: same here, lime_sdrdevice just returns lime_SPI handle, everything regarding lime_SPI should be in it's own files.

src/include/limesuiteng/types_c.h
```diff
@@ -0,0 +1,63 @@
...
+#include <stdbool.h>
+
+#if defined(_WIN32)
+    #if defined(LIME_C_BUILD)
+        #define LIME_C_API __declspec(dllexport)
+    #else
+        #define LIME_C_API __declspec(dllimport)
+    #endif
+#elif defined(__GNUC__)
+    #define LIME_C_API __attribute__((visibility("default")))
```
rjonaitis: import/export definitions should be in a dedicated header.

src/include/limesuiteng/SDRDevice.cpp
```diff
@@ -1,11 +1,40 @@
 #include "limesuiteng/SDRDevice.h"
+#include "limesuiteng/SDRDescriptor.h"
 #include "limesuiteng/StreamConfig.h"
 
 #include "limesuiteng/Logger.h"
 
 using namespace lime;
 using namespace std::literals::string_literals;
 
+OpStatus SDRDevice::SetAntenna(ChannelId c, std::string_view name)
```
rjonaitis: No need for introducing "ChannelId", it's not used anywhere else, just adds complexity.
These Set/Get antenna by name functionality could have been implemented directly within lime_sdrdevice_set/get_antenna.

rjonaitis (commented):
No need for filename suffixes "_c", especially for the intended public C headers.
```
#include <limesuiteng/limesuiteng_c.h>
#include <limesuiteng/rfstream_c.h>
```
That just looks stupid for library that will provide only C API.

same for src/lib files, everything in there will be related to the C interface implemenation.

.gitattributes
rjonaitis: Fine, it can stay.

src/CMakeLists.txt
dmitrymmm: Done in a3c7dd89: src/lib has its own CMakeLists.txt managing the wrapper sources; src/CMakeLists.txt just add_subdirectory(lib).

src/include/limesuiteng/sdrdevice_c.h
dmitrymmm: Done in a3c7dd89: lime_GPIO and its functions moved to gpio_c.h + src/lib/gpio_c.cpp; sdrdevice_c.h keeps only lime_sdrdevice_get_gpio().

src/include/limesuiteng/sdrdevice_c.h
dmitrymmm: Done in a3c7dd89: spi_c.h + src/lib/spi_c.cpp; sdrdevice_c.h keeps only lime_sdrdevice_get_spi().

src/include/limesuiteng/types_c.h
dmitrymmm: Done in a3c7dd89: LIME_C_API moved to a dedicated config_c.h (mirroring the C++ config.h); types_c.h includes it.

src/include/limesuiteng/SDRDevice.cpp
dmitrymmm: Done in a3c7dd89: ChannelId is removed entirely (types.h and both SDRDevice methods). The name<->index resolution now lives directly in lime_sdrdevice_set_antenna, and the symmetric lime_sdrdevice_get_antenna is added, returning the current antenna name. Unit tests retarget the C entry points against a stub device (tests/api/antenna_c_tests.cpp).

dmitrymmm:
Re the `_c` filename suffixes: agreed for src/lib - dropped in e5dd0726 (device_registry.cpp, sdrdevice.cpp, gpio.cpp, spi.cpp, rfstream.cpp, sdrdescriptor.cpp, versioninfo.cpp; capi_private.h -> private.h).

For the public headers the clean names are currently occupied by the transitional C++ headers in the same directory: `types.h` and `config.h` collide exactly, and `sdrdevice.h` vs `SDRDevice.h`, `rfstream.h` vs `RFStream.h` etc. collide on case-insensitive filesystems (Windows/macOS checkouts). Plan: the C headers take the clean names in one mechanical rename when the C++ headers leave the public include directory at the freeze. If you'd rather not wait, the alternative is non-colliding names now (`device.h`, `stream.h`, `descriptor.h`, `version.h`, `common.h`, `export.h`) - happy to do either.

dmitrymmm:
Windows build fixed in 2a10a864. Root cause: the windows-latest job builds static (BUILD_SHARED_LIBS=OFF), and LIME_C_API treated _WIN32 unconditionally as a DLL build - consumer TUs declared the lime_* entry points dllimport and referenced __imp_ symbols a static lib doesn't provide (the LNK2019 in antenna_c_tests.obj). config_c.h now follows the same LIME_DLL / LIME_DLL_EXPORTS scheme as the C++ LIME_API: export when building shared, import when consuming, empty for static. Verified locally in both configurations (shared: 33 symbols exported, external C99 consumer links and runs; static: gtest-runner links and 50/50 tests pass).

src/include/limesuiteng/config_c.h
rjonaitis, line 1: This header essentially duplicates what's in limesuiteng/config.h.
You can just reuse LIME_API from it, no need to have separate definition specific to C exports.

rjonaitis (commented):
To solve the headers name collisions, rename colliding C++ headers from .h to .hpp

src/include/limesuiteng/config_c.h
dmitrymmm: Done in 6cc05b3b: config_c.h is removed and the C headers include limesuiteng/config.h and mark entry points with LIME_API directly (config.h is plain preprocessor, so it's C-clean). LIME_C_API is gone from the headers and docs/conf.py. Verified shared (all 33 lime_* symbols exported, 50/50 tests) and static (links, 50/50 tests) builds.

dmitrymmm:
Done in aba4a539: colliding C++ headers renamed from .h to .hpp (types, SDRDevice, SDRDescriptor, RFStream, and the generated VersionInfo), and the public C headers take the clean names with the _c suffix dropped (types.h, device_registry.h, sdrdevice.h, sdrdescriptor.h, rfstream.h, versioninfo.h, gpio.h, spi.h). config.h stays shared between both APIs as the LIME_API home. All includes across src, cli, GUI, plugins, tests, docs and the conda recipes are updated. The C header include guards keep the _C_H suffix on purpose: renaming them would collide with the C++ guards (e.g. LIMESUITENG_TYPES_H in types.hpp) and silently skip declarations in translation units that include both.

Code review of the branch was performed as part of this change; eef9a2dd rewords comments to plain ASCII prose and removes a stale comment that referenced the removed ChannelId overloads.

Verified: shared and static builds compile, 50/50 unit tests pass in both, all 33 lime_* symbols remain exported, and every public C header compiles standalone with gcc -std=c99 -Wall -Wextra -Werror -pedantic.

dmitrymmm:
Formatting CI failures fixed in f0df1419. Two spots from the header rename commit tripped the checks: clang-format wanted the lime_spi_transact declaration in spi.h joined onto a single line (it fits the 132 column limit), and cmake-format wanted the VersionInfo.hpp configure_file call in src/include/CMakeLists.txt wrapped, because the longer hpp extension pushed the line past the same limit; it now matches the wrapped style of the VersionInfo.cpp call next to it. Verified locally with the same tools the workflow runs (git-clang-format against the PR base reports no changes for the flagged files, cmake-format --check passes), plus a rebuild and 50/50 unit tests.

rjonaitis:
The 'conda package' failure can be ignored here, misconfiguration of CI, it runs with the updated yaml files with renamed headers filenames, but checkouts base branch revision, therefore tests fail due to header files not found.

PR #211, opened 2026-07-15
Add board check script and runbook

dmitrymmm:
Adds tools/board-check: a short script that verifies a connected board end to end with the shipped CLI tools (enumeration, identity, configuration, streaming with the chip test signal, so no antennas needed), plus a runbook with prerequisites, failure triage and the release candidate check procedure. Stops at the first failing step, prints PASS or FAIL, exit code 0 only on pass.

Fixes #192.

### DIFF
--- a/tools/board-check/board_check.sh
+++ b/tools/board-check/board_check.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# LimeSuiteNG board check. Runs each step in order and stops at the
+# first failure, so the last printed step is the one that failed.
+# Exit code 0 means the board passed. See RUNBOOK.md for details.
+
+set -e
+trap 'echo "FAIL"' ERR
+
+echo "1. Board enumerates"
+limeDevice
+
+echo "2. Identity readable"
+limeDevice --full
+
+echo "3. Configuration accepted: 10 MHz sample rate, 1 GHz LO, Rx test signal"
+limeConfig --initialize --samplerate 10e6 --rxen 1 --rxlo 1e9 --rxtestsignal 1
+
+echo "4. Rx streaming works: 2 seconds through the whole digital path"
+limeTRX --time 2000
+
+echo "PASS"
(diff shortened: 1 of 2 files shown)


tools/board-check/board_check.sh
```diff
@@ -0,0 +1,21 @@
...
+trap 'echo "FAIL"' ERR
+
+echo "1. Board enumerates"
+limeDevice
+
+echo "2. Identity readable"
+limeDevice --full
+
+echo "3. Configuration accepted: 10 MHz sample rate, 1 GHz LO, Rx test signal"
+limeConfig --initialize --samplerate 10e6 --rxen 1 --rxlo 1e9 --rxtestsignal 1
```
rjonaitis, line 16: Unreliable, device configuration could fail simply due to provided parameter being out of supported range. Device health checking needs to be done with device specific parameters.
There is 'limeOEM --test` cli, that suppose to run board health checks and hardware components tests.

tools/board-check/board_check.sh
```diff
@@ -0,0 +1,21 @@
...
+
+echo "4. Rx streaming works: 2 seconds through the whole digital path"
+limeTRX --time 2000
```
rjonaitis, line 19: unreliable, would only fail if stream confiuration would be invalid. If no samples would be received over the 2 seconds, it would still return success.
From end user perspective streaming tests are part of 'limeOEM'
From development perspective there is a set of streaming "unit tests" that can be run with actual hardware https://github.com/myriadrf/LimeSuiteNG/blob/f0290a4375a01c9bc571b5acf955915f4c4baca0/tests/CMakeLists.txt#L61-L71

rjonaitis:
The submited work is not verified, and has existing alternatives, Please defer work that requires actual hardware to run until you get the device and be able to verify the process.

dmitrymmm:
Agreed on all points - limeOEM --test and gtest-runner-hardware already cover this, and my pass criteria were not reliable: limeTRX exit code does not prove samples actually flowed. Closing this PR. I will redo #192 as a runbook around the existing tools, with device specific parameters, once there is a board to verify against.

PR #214, opened 2026-07-16
Add ABI compatibility check to CI

dmitrymmm:
Adds an abidiff gate to pull requests: the library is built from the PR and from its base branch, and incompatible ABI changes fail the check. Compatible additions pass, matching the append only growth rule. This is the abidiff half of issue #190; the v1.0.0-rc1 tag half stays with the maintainer, and the baseline can be switched to the tag once it exists.

Verification in this PR: the green run below shows the pass path; a deliberate ABI break will be pushed and reverted to show the fail path in the check history.

Part of #190.

### DIFF
--- a/.github/workflows/abi-check.yml
+++ b/.github/workflows/abi-check.yml
@@ -22,13 +22,13 @@ jobs:
 
       - name: Build pull request library
         run: |
-          cmake -B build-head -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_CLI=OFF -DBUILD_GUI=OFF -DBUILD_PLUGINS=OFF -DBUILD_DRIVERS=OFF -DBUILD_DOCS=OFF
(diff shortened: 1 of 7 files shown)


.github/workflows/abi-check.yml
```diff
@@ -0,0 +1,52 @@
...
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -qq abigail-tools
+          sudo ./install_dependencies.sh
+
+      - name: Build pull request library
+        run: |
+          cmake -B build-head -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_CLI=OFF -DBUILD_GUI=OFF -DBUILD_PLUGINS=OFF -DBUILD_DRIVERS=OFF -DBUILD_DOCS=OFF -DBUILD_EXAMPLES=OFF
+          cmake --build build-head --parallel 4
```
rjonaitis: Only the 'liblimesuiteng' needs to be built, you can specify ` --target limesuiteng `, so it won't build anything else or require action changes if more components would be added, that would need disabling (Like it was with the examples missing dependency).

dmitrymmm: Done in a68ab594, the disable flags are dropped and both builds now use --target limesuiteng.

Issue #149, opened 2025-09-27
Lack of Windows executables


G0MJW:
Lime's documentation says to use LimesuiteNG but there don't seem to be any windows executables?

9600:
Which LimeSDR board are you using? If it's LimeSDR XTRX, just to note that the PCIe driver has not been ported to Windows yet, so it's only supported on Linux at present.

G0MJW:
I got an original Lime Mini 1.2e from a silent key sale – unfortunately, it appears it is dead. As I could not speak to it in Windows I managed to load the programmes into Linux Mint and it was equally uncommunicative but more helpful error messages, basically saying no data received by the USB chip from the FPGA. I then tried to speak to the FPGA it with JTAG but without success. No data from the FPGA, the chip could not be identified. 

 

Mike

9600:
Thanks. 

With the original LimeSDR Mini you can use either classic Lime Suite or Lime Suite NG. Though the former will be better tested with that hardware, as the focus with NG has been mostly on more recent boards and so Mini 1.0 won't be as well tested.

With classic Lime Suite you can run LimeQuickTest to test the hardware. The best place for debugging such issues is the forums and if perhaps you could post the output there.

https://discourse.myriadrf.org

Then a colleague can take a look.

dmitrymmm:
PR #217 makes the Windows CI publish the built binaries (limesuiteng, legacy API, CLI tools, x64 and Win32) as downloadable workflow artifacts. Proper installers are planned as part of the 1.0 release packaging.

PR #217, opened 2026-07-16
Upload Windows build binaries as CI artifacts

dmitrymmm:
The Windows CI job builds limesuiteng.dll, the legacy API, limeGUI and all the CLI tools for x64 and Win32, but nothing was published. Stage the install into a folder and upload it as a workflow artifact, so every run provides downloadable Windows binaries. Proper user facing installers remain part of the release packaging tasks.

Verified: downloaded the limesuiteng-windows-x64 artifact from this PR run and ran the tools on a Windows 11 machine, limeDevice.exe executes and reports no devices as expected without hardware. The artifact also includes limeGUI.exe with the wxWidgets runtime DLLs.

Part of #149

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -77,6 +77,15 @@ jobs:
       - name: Install
         run: cmake --build ${{github.workspace}}/build --config ${{env.BUILD_TYPE}} --parallel ${{env.CMAKE_BUILD_PARALLEL_LEVEL}} --target install
 
+      - name: Stage install for artifact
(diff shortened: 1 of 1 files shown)


rjonaitis:
> The GUI is not included because the runner has no wxWidgets

It is included. For Windows builds CMake downloads wxWidgets prebuilt libraries.
https://github.com/myriadrf/LimeSuiteNG/blob/e113e7f19a51b97be0dcac8ff35f6e1258c90c13/GUI/CMakeLists.txt#L24-L72

Issue #140, opened 2025-07-11
Driver installation error


vd2org:
Build from source fails on the driver installation part.

Branch: develop

Logs:

```shell
modinfo: ERROR: Module limepcie not found.
```

osqzss:
Same here. I'm also unable to install the `limepcie` driver on Ubuntu 24.04 LTS when building from source for the LimeSDR XTRX.
Has anyone found a workaround or solution for this issue?

rjonaitis:
Does that happen with a fresh clean build?

Might be the same issue as https://github.com/myriadrf/LimeSuiteNG/issues/134, first time the cmake is configured it caches the location of kernel headers/instal path. Ubuntu sometimes auto updates it's kernel to a new version, after that if limesuiteng cmake install is executed it tries to install into the last cached location, which is no longer used by the active kernel, resulting in module not found error.
To fix it, the "build" directory has to be deleted, so that cmake would reinitialize with a new kernel location.

osqzss:
I was able to install the limepcie module after deleting the build directory and re-running CMake, just as you suggested. 
I also noticed that Secure Boot needs to be disabled in the BIOS for the installation to succeed.
Thank you very much for your helpful advice.

rjonaitis:
"Secure boot" requires kernel modules to be digitally signed with a trusted key.

If linux DKMS is installed on the system, LimeSuiteNG will detect it and can use it for building the kernel modules. The DKMS will generate a self certified key and sign built modules, they would be able to be loaded when secure boot is enabled

dmitrymmm:
PR #219 makes this failure mode self explanatory: when the kernel was updated after CMake configuration, the install now stops with a clear message to delete the build directory and reconfigure.

PR #219, opened 2026-07-16
limepcie: fail install clearly when the kernel changed after configure

### LINKED ISSUE #134: LimePCIe kernel module installs to cached old location after system kernel update
CMake caches directory of kernel version that was first used for project configuration step. If not using DKMS, and system kernel is updated causing version change, cmake install procedure installs built module to the cached location instead of a new kernel directory.

Currently CMake cache removal is required to auto refresh the kernel location.

dmitrymmm:
The kernel release was captured at CMake configure time. After a system kernel update the module was still built and installed for the old kernel, and the running kernel could not find it, failing late with a confusing modinfo error (#140).

Reworked after review feedback:

- Auto detected kernel release: a kernel.release stamp file is regenerated whenever the running kernel changes and is a dependency of the module object, so the module gets rebuilt for the currently active kernel. The build resolves the kernel directory at build time through the generated Makefile, and the install destination is resolved at install time. The regular install flow (cmake --build --target install) builds first, so after a kernel update everything self repairs. A plain cmake --install of a module built for another kernel is refused with instructions to rebuild.
- Manually set KMOD_KERNEL_RELEASE (cross compiling, packaging): used exactly as before, no uname involvement anywhere.
- DKMS path unchanged.

Verified on CI runners in four scenarios: matching kernel builds and installs normally; a kernel change between configure and install triggers the automatic rebuild (vermagic checked) and the module lands in the new kernel directory; plain cmake --install of a stale module is refused with a clear message; a manually set kernel release is used as is, with no install time resolution in the generated install rules. Run: https://github.com/dmitrymmm/LimeSuiteNG/actions/runs/29582586239

Fixes #134
Related #140

### DIFF
--- a/drivers/linux/limepcie/cmake/add_kernel_module.cmake
+++ b/drivers/linux/limepcie/cmake/add_kernel_module.cmake
@@ -40,10 +40,9 @@ function(add_kernel_module)
     set(MODULE_KOBJECT ${KBUILD_FILE_DIR}/${KMOD_NAME}.ko)
     set(KERNEL_SOURCE_DIR /lib/modules/${KMOD_KERNEL_RELEASE}/build)
     if(KMOD_KERNEL_AUTODETECTED)
-        # build through the generated Makefile, it resolves the kernel directory at
-        # build time, so the module is always built for the currently running kernel
-        set(KBUILD_CMD $(MAKE) -C ${KBUILD_FILE_DIR})
-        set(KBUILD_CLEAN_CMD $(MAKE) -C ${KBUILD_FILE_DIR} clean)
+        # Native build for the running kernel; kbuild detects ARCH on its own.
+        set(KBUILD_CMD $(MAKE) -C ${KERNEL_SOURCE_DIR} M=${KBUILD_FILE_DIR} modules)
+        set(KBUILD_CLEAN_CMD $(MAKE) -C ${KERNEL_SOURCE_DIR} M=${KBUILD_FILE_DIR} clean)
     else()
         set(KBUILD_CMD
             $(MAKE)
--- a/drivers/linux/limepcie/cmake/add_kernel_module.cmake
+++ b/drivers/linux/limepcie/cmake/add_kernel_module.cmake
@@ -26,6 +26,10 @@ function(add_kernel_module)
             COMMAND uname -r
             OUTPUT_VARIABLE KMOD_KERNEL_RELEASE
             OUTPUT_STRIP_TRAILING_WHITESPACE)
+        # the value comes from the running system, not from the user, so it can go
+        # stale when the kernel gets updated after CMake configuration
+        set(KMOD_KERNEL_AUTODETECTED TRUE)
+        set(KMOD_KERNEL_AUTODETECTED TRUE PARENT_SCOPE)
     endif()
 
     # where Kbuild file will be placed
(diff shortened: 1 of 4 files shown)


rjonaitis:
Not sure yet, if this wouldn't cause problems with cross-compiling and it's packaging. As the target platform kernel version would differ from the hosts current kernel.
The actual fix would be to check if the KERNELHEADERS_DIR and KMOD_KERNEL_RELEASE have been set manually, that would indicate cross compiling, and just use those values.
If they were auto detected, then a prebuild step check should be made, to update those values with the currently active kernel.

There could be a custom_target or a custom_command to generate a file with the kernel/header version, and add that file as a dependency to trigger rebuild of the driver if kernel changes.
https://github.com/myriadrf/LimeSuiteNG/blob/3c3f6d04bb9cc18f4faa939f2d3f269653446843/drivers/linux/limepcie/cmake/add_kernel_module.cmake#L87
https://github.com/myriadrf/LimeSuiteNG/blob/3c3f6d04bb9cc18f4faa939f2d3f269653446843/drivers/linux/limepcie/cmake/add_kernel_module.cmake#L58

dmitrymmm:
The check here only runs when KMOD_KERNEL_RELEASE was auto detected, manually set values are used as is. But agreed, packaging in a chroot would still hit it since the host uname differs from the target benignly, and failing is worse than self repair anyway. I will rework it your way: manually set KERNELHEADERS_DIR/KMOD_KERNEL_RELEASE are trusted without any uname involvement, and in the auto detected case a generated kernel version stamp file becomes a dependency of the module object, so a kernel update triggers a rebuild for the active kernel and the install follows it.

dmitrymmm:
Reworked as suggested: the auto detected case regenerates a kernel.release stamp that triggers the module rebuild for the active kernel, the build and install resolve the kernel directory at build and install time, and a manually set KMOD_KERNEL_RELEASE is used as is without any uname involvement. Verified in four runner scenarios, including the automatic rebuild after a kernel change (vermagic checked) and the manual value staying untouched: https://github.com/dmitrymmm/LimeSuiteNG/actions/runs/29582586239

drivers/linux/limepcie/cmake/add_kernel_module.cmake
```diff
@@ -26,6 +26,10 @@ function(add_kernel_module)
             COMMAND uname -r
             OUTPUT_VARIABLE KMOD_KERNEL_RELEASE
             OUTPUT_STRIP_TRAILING_WHITESPACE)
+        # the value comes from the running system, not from the user, so it can go
+        # stale when the kernel gets updated after CMake configuration
+        set(KMOD_KERNEL_AUTODETECTED TRUE)
+        set(KMOD_KERNEL_AUTODETECTED TRUE PARENT_SCOPE)
```
rjonaitis: There is going to be multiple linux drivers, in parallel directories using this 'add_kernel_module' script. Modifying variable at parent scope might clash, it would still work as both directories would use same values. Cleaner solution would be to move this kernel version related check and variables to the top level of the entire project, so there would be a single common target responsible for handling kernel versioning, and the values could be passed down to all targets that would need it.

dmitrymmm:
Agreed, the parent-scope write is fragile. Plan: resolve the kernel release once at the drivers/linux level, before any driver subdirectory. Shared through global properties plus a single kernel.release stamp target that every module depends on, so parallel drivers can't clash and nothing writes to parent scope. Manually set KMOD_KERNEL_RELEASE / KERNELHEADERS_DIR stay trusted as-is, no uname involved. Rework incoming.

drivers/linux/limepcie/cmake/add_kernel_module.cmake
```diff
@@ -39,17 +39,24 @@ function(add_kernel_module)
...
-        # Informs kbuild that an external module is being built.
-        # The value given to "M" is the absolute path of the directory where the external module (kbuild file) is located.
-        M=${KBUILD_FILE_DIR}
-        modules)
-
-    set(KBUILD_CLEAN_CMD $(MAKE) -C ${KERNEL_SOURCE_DIR} ARCH=${KMOD_ARCH} M=${KBUILD_FILE_DIR} clean)
+    if(KMOD_KERNEL_AUTODETECTED)
+        # build through the generated Makefile, it resolves the kernel directory at
+        # build time, so the module is always built for the currently running kernel
+        set(KBUILD_CMD $(MAKE) -C ${KBUILD_FILE_DIR})
```
rjonaitis: -C ${KBUILD_FILE_DIR} ends up triggering the "default" Makefile https://github.com/myriadrf/LimeSuiteNG/blob/8ea99e204ab7bce68e7d197cfbfe5cfc0f8ab1b0/drivers/linux/limepcie/cmake/add_kernel_module.cmake#L127-L138 that is generated for standalone use when module source is installed into system directory "/usr/src", and the build is done manually with just 'make' by the user for the host running kernel.
To avoid any confusion and parameters mismatch, that Makefile should not be called by CMake.

CMake should call kernel's Makefile with explicit arguments "-C ${KERNEL_SOURCE_DIR} M=${KBUILD_FILE_DIR}"

drivers/linux/limepcie/cmake/add_kernel_module.cmake
```diff
@@ -39,17 +39,27 @@ function(add_kernel_module)
...
-        modules)
-
-    set(KBUILD_CLEAN_CMD $(MAKE) -C ${KERNEL_SOURCE_DIR} ARCH=${KMOD_ARCH} M=${KBUILD_FILE_DIR} clean)
+    if(KMOD_KERNEL_AUTODETECTED)
+        # Call the kernel's Makefile directly, resolving its directory from the
+        # running kernel at build time, so a kernel change (tracked by the
+        # kernel-release-stamp dependency) rebuilds the module for the new kernel.
+        # The generated Makefile is only a convenience for a manual build from
+        # /usr/src and must not be invoked by CMake.
+        set(KBUILD_CMD sh -c "make -C /lib/modules/`uname -r`/build M=${KBUILD_FILE_DIR} modules")
```
rjonaitis: another indirection through shell, with inline kernel directory change.  All the variables are already set up by CMake, you just need to combine them into comand, just don't need the ARCH for native build:
https://github.com/myriadrf/LimeSuiteNG/blob/8ea99e204ab7bce68e7d197cfbfe5cfc0f8ab1b0/drivers/linux/limepcie/cmake/add_kernel_module.cmake#L42-L50

dmitrymmm:
Makes sense. Dropped the shell wrapper and just combined the CMake variables: `$(MAKE) -C ${KERNEL_SOURCE_DIR} M=${KBUILD_FILE_DIR} modules`, with ARCH left to kbuild for the native build. The generated Makefile stays for manual /usr/src use. Pushed.

PR #223, opened 2026-07-17
limepcie: restore control channel serialization

dmitrymmm:
The control semaphore was initialized but all its uses were commented out, so two processes could interleave writes and reads on the control CSR and corrupt each others commands. The semaphore is taken again for the whole command, with EBUSY returned when it cannot be acquired within the callers timeout, released on every exit path, and the status poll loop gets cpu_relax.

Verified on runners: the module builds, installs, loads and unloads cleanly on ubuntu-22.04 and ubuntu-24.04 kernels (https://github.com/dmitrymmm/LimeSuiteNG/actions/runs/29584464440). Concurrency behaviour under a real device still deserves a stress run with hardware.

Addresses the disabled lock and busy wait items of #196. The no-op LOCK ioctl is left for separate work, it needs per-open file state for crash safe lock release.


rjonaitis:
Update driver patch version https://github.com/myriadrf/LimeSuiteNG/blob/develop/drivers/linux/limepcie/VERSION

dmitrymmm:
Done, bumped to 0.1.11.

PR #228, opened 2026-07-20
soapy: create the hardware stream on activate, not setup

### LINKED ISSUE #127: LimeSuiteNG soapy stream not starting with certain applications
Lubuntu 24.04 with latest LimeSuiteNG using a Lime Mini v2 as of this morning. 
Soapy Lib Version: v0.8.1-4build1
API Version: v0.8.0
ABI Version: v0.8

I recall something very similar happening to a fork of both the Airspy and AirspyHF soapy plugins where in some applications due to whatever reason, the stream didn't start fast enough? 

Rechecking CubicSDR soapy results. It would appear the stream never starts.

```
[INFO] Make connection: 'LimeSDR Mini xxxxxxx'
[INFO] Reference clock 40.00 MHz
[INFO] Device name: LimeSDR-Mini_v2
[INFO] Reference: 40 MHz
[INFO] LMS7002M register cache: Disabled
SDR thread starting.
device init()
[INFO] Sampling rate set(2.100 MHz): CGEN:268.800 MHz, Decim: 2^5, Interp: 2^5
[INFO] Rx LPF configured
[ERROR] LoadConfig() - file not found
[INFO] Sampling rate set(2.100 MHz): CGEN:268.800 MHz, Decim: 2^5, Interp: 2^5
[INFO] Sampling rate set(2.100 MHz): CGEN:268.800 MHz, Decim: 2^5, Interp: 2^5
SDRThread : Device Stream set to MTU: 8160
starting readLoop()
```

Rechecking GQRX soapy results. Also appears the stream does not start.
```
[INFO] Make connection: 'LimeSDR Mini xxxxxx'
[INFO] Reference clock 40.00 MHz
[INFO] Device name: LimeSDR-Mini_v2
[INFO] Reference: 40 MHz
[INFO] LMS7002M register cache: Disabled
[INFO] Sampling rate set(10.003 MHz): CGEN:320.104 MHz, Decim: 2^3, Interp: 2^3
[INFO] Rx LPF configured
[INFO] Sampling rate set(5.000 MHz): CGEN:640.000 MHz, Decim: 2^5, Interp: 2^5
[INFO] Rx LPF configured
[ERROR] No samples or timeout
[ERROR] No samples or timeout
[ERROR] No samples or timeout
[ERROR] No samples or timeout
[ERROR] No samples or timeout
[ERROR] No samples or timeout
```

SigDigger. Stream starts and works. 
SDRAngel with Soapy. Stream starts and works.

dmitrymmm:
Fixes #127. Stream creation moved from setupStream() to the first activateStream(), so apps that configure the sample rate after stream setup get a stream built from the final configuration. Both directions still share one hardware stream, accumulated across setups before activation. Null guards added so close/read/write before activation return an error instead of crashing.

Verified: plugin builds clean against SoapySDR 0.8.1, full unit test suite passes.

### DIFF
--- a/plugins/soapysdr/Streaming.cpp
+++ b/plugins/soapysdr/Streaming.cpp
@@ -176,8 +176,18 @@ SoapySDR::Stream* Soapy_limesuiteng::setupStream(
     //     config.bufferLength = std::stoul(args.at("bufferLength"));
     // }
 
-    // Stream creation is deferred to activateStream(), the device may not be
-    // fully configured (e.g. sample rate) at this point
+    // Setup is deliberately split from start so that activateStream() stays fast.
+    // The sample rate only hints buffering sizes, so when the app has not
+    // configured it yet, fall back to the device's current rate.
+    if (sampleRate[SOAPY_SDR_RX] <= 0.0)
+        sampleRate[SOAPY_SDR_RX] = sdrDevice->GetSampleRate(0, TRXDir::Rx, 0);
+    config.hintSampleRate = sampleRate[SOAPY_SDR_RX];
+
+    rfstream = sdrDevice->StreamCreate(config, 0);
+    if (!rfstream)
+    {
+        throw std::runtime_error("Soapy_limesuiteng::setupStream() failed: " + std::string(GetLastErrorMessage()));
+    }
 
     auto samplesPerPacket = config.linkFormat == DataFormat::I16 ? 1020 : 1360;
     // TODO: figure out a way to actually get the correct packet per batch count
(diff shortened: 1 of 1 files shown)


rjonaitis:
Stream setup takes relatively long time, that's why it's split into setup(), and start() steps, so that the start could be as fast as possible and have more consistent timing.
Technically stream can be set up regardless of sampling rate. Currently sampling rate is used just as a hint, to chose the total amount of buffering and transfer batch sizes to maintain adequate latency. If sampling rate hint is not provided for stream setup, then the currently set sample rate is readback.

dmitrymmm:
Good point, reworked: setupStream keeps creating the stream so start() stays fast, and uses the device's current rate as the hint when the app has not set one. The stream is recreated at activate only if the rate changed since setup, which covers the late-configuring apps from #127.

PR #234, opened 2026-07-20
packaging: produce Windows zip and NSIS installer in CI

dmitrymmm:
Part of #149. CI already uploads raw Windows binaries (#217); this adds a versioned zip and an NSIS installer built by CPack and uploaded as artifacts for x64 and Win32: component chooser, optional PATH entry for the CLI tools, start menu shortcut for limeGUI, uninstall before reinstall. Publishing them as release assets can follow once v1.0.0-rc1 is tagged.

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -86,6 +86,17 @@ jobs:
           name: limesuiteng-windows-${{matrix.arch}}
           path: ${{github.workspace}}/artifact
 
+      - name: Package zip and installer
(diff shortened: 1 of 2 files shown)


rjonaitis:
Did you test the installer? Does it take care of C++ redistributable runtime libraries?

dmitrymmm:
Runtime DLLs are bundled now (InstallRequiredSystemLibraries). The installer itself is only CI-built so far, I'll run the artifact on a Windows machine and report back.

PR #318, opened 2026-07-23
lms7002m: bound the calibration wait loops with a timeout

dmitrymmm:
Fixes #309.

The Rx DC auto-cal poll on register `0x05C1` and the two RP-bias convergence loops had no iteration limit, so a stuck engine or a bad SPI read would hang the caller forever. This caps the DC poll and caps each RP-bias sweep at the field's 32-value range, restoring `MUX_BIAS_OUT` and returning an error if it fails to converge. Built locally.

---


embedded/lms7002m/calibrations.c
```diff
@@ -501,8 +501,15 @@ static void lms7002m_calibrate_rx_dc_auto(lms7002m_context* self)
     }
 
     {
+        int timeout = 1000;
```
rjonaitis: At this moment it's unclear how many retries should be enough.
The calibration algorithm is running inside the chip, it should complete in fixed number of cycles, the cycles rate is dependent on reference clock.
The retry rate is affected by the SPI bus frequency. SPI frequency is unknown, as it's set by the device's hardware.
1000 retries might be enough, when executing from software, as it's slow, but migth not be enough when executing within actual embedded system.
Need to estimate the worst case scenario by maximum chip's supported SPI frequency, and the lowest supported reference clock.
The algorithm duration is not provided in the data sheet, so will need to be measured.

dmitrymmm:
You're right, and thanks - a retry count is the wrong unit here. It is a proxy for time whose meaning changes with the SPI bus speed, and I had no basis for picking 1000. I have reworked it so that neither loop depends on that guess:

- **Rx DC auto-calibration poll (0x05C1)** now waits in *time*, sleeping between reads, so the limit no longer moves with bus speed. I left it at a deliberately generous 1 s and named the constant: it is a safety net that turns an endless hang into a reported error, not an estimate of how long the algorithm takes. It should be tightened once the duration is actually measured on hardware, which I cannot do without a board.
- **RP bias sweeps** no longer use a count at all. They are now bounded by the range of `RP_CALIB_BIAS` itself, a 5 bit field, so a sweep cannot need more steps than the field has values. That is the same guard style as `lms7002m_calibrate_internal_adc`, and it carries no timing assumption whatsoever. It also fixes a real defect: `rpCalib` is `uint16_t`, so `--rpCalib` at 0 wrapped around through the field and that loop could never terminate.

If you would rather the DC poll were structural too, the alternative I see is to fail only when the chip reports the engine idle without the done flag - happy to go that way instead.

PR #319, opened 2026-07-23
limepcie: bail out if PCI configuration fails during probe

dmitrymmm:
Fixes #308.

`limepcie_device_init` ignored the return of `limepcie_configure_pci` and fell through to the reset write and `TestCSR`, which dereference `bar0_addr`. On a configure failure `bar0_addr` is NULL, so probe would oops instead of failing cleanly. Return the error immediately. (Kernel module; change is a guarded early return.)

---


rjonaitis:
Already fixed.

